PR #28, opened 2015-05-21
Enable GC option

dantleech:
This PR adds an option to enable garbage collection, and disables it by default.


lib/Console/Application.php
```diff
@@ -34,6 +34,8 @@ public function __construct(Configuration $configuration = null)
         $this->add(new RunCommand());
         $this->add(new ReportCommand());
 
+        gc_disable();
```
dantleech: Move to the run command? Does not make sense to disable this for the report generation command.

PR #50, opened 2015-05-25
Revs

dantleech:
Support for revolutions within iterations.


tests/Unit/Result/Dumper/XmlDumperTest.php
```diff
@@ -44,7 +44,7 @@ public function testDump()
             <parameter name="one" value="two"/>
             <parameter name="three" value="four"/>
           </parameter>
-          <iteration time="100"/>
+          <iteration repeats="1" time="100"/>
```
dantleech: s/repeats/revs

PR #62, opened 2015-06-04
Refactoring reports

dantleech:
This PR aims to refactor reports to enable them to operate only groups of tables, for instance to create a summary report.

````
$config->addReport(array(
    'name' => 'console_table',
    'groups' => array('some_group_name'), // tables can be aggregated from multiple benchmarks
    'aggregate' => 'table',
    'deviation' => true,
    'order' => array('rps' => 'desc'),
    'rps' => true
));
````

### DIFF
--- a/lib/Benchmark/Runner.php
+++ b/lib/Benchmark/Runner.php
@@ -142,6 +142,7 @@ class Runner
         $subjectResult = new SubjectResult(
             $subject->getMethodName(),
             $subject->getDescription(),
+            $subject->getGroups(),
             $iterationsResults
         );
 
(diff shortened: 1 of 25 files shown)

### CODE REVIEW

File: lib/Benchmark/Runner.php [hunk 1]
```diff
@@ -187,10 +188,11 @@ private function runIterationsSeparateProcess(Benchmark $benchmark, Subject $sub
 
         for ($index = 0; $index < $iterationCount; $index++) {
             $command = sprintf(
-                'php %s run %s --subject=%s --nosetup --dump --parameters=%s --iterations=%d --processisolation=none',
+                'php %s run %s --subject=%s --subject=%s --nosetup --dump --parameters=%s --iterations=%d --processisolation=none',
```
dantleech: Revert

PR #64, opened 2015-06-06
Initial aggregate subject

### LINKED ISSUE #17: Generate bench mark summary
The benchmark summary should list the aggregated results for each benchmark in a single table, showing:

- avg, min, max and total time
- percentage compared to the minimum time (e.g. +50% faster than subject A)

dantleech:
This PR will introduce support for aggregating subjects into a single table.

Fixes #17


examples/phpbench
```diff
@@ -5,7 +5,7 @@ require_once(__DIR__ . '/../vendor/autoload.php');
 $config = new PhpBench\Configuration();
 $config->addReport(array(
     'name' => 'console_table',
-    'aggregate' => 'iteration',
+    'aggregate' => 'run',
```
dantleech: + BC break

PR #85, opened 2015-06-17
Default report

dantleech:
New `simple_table` report which is now used by default.

NOTE: This is not a very nice way of implemting this, should refactor it in 0.4


lib/Report/Generator/ConsoleSimpleReportGenerator.php
```diff
@@ -0,0 +1,50 @@
...
+use Symfony\Component\Console\Formatter\OutputFormatterStyle;
+use DTL\Cellular\Workspace;
+use PhpBench\Console\Output\OutputIndentDecorator;
+use Symfony\Component\Console\Helper\TableHelper;
+use PhpBench\Report\Generator\ConsoleTableReportGenerator;
+use PhpBench\Result\SuiteResult;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+use PhpBench\OptionsResolver\OptionsResolver as BCOptionsResolver;
+
+
```
dantleech: EL

PR #96, opened 2015-06-27
Refactoring aggregation

dantleech:
This PR intends to improve the aggregation logic and support "exploding" aggregated tables into separate tables.


dantleech:
Reports are becoming to complicated, going to try a completely different approach.

PR #98, opened 2015-07-07
Experimenting with using SQLite

dantleech:
Use in-memory SQLite to prepare data for reports, and then delegate the rendering to output formatters.

The aims are:

- Provide a familiar SQL language interface for data preparation.
- Remove dependence on the Cellular libarary (if it is no longer required).
- Allow definition of named reports, which could be composite
- .. allowing the generation of full HTML reports.


dantleech:
See last commit message.

PR #101, opened 2015-08-01
Changing to ansible-like iterators

dantleech:
This PR aims to:

- Enable the definition of arbitrary rows
- Enable rows to be added dynamically (e.g. from the result of an XPath query or from a list of parameters
- Enable parameterable queries (e.g. `./iteration[parameter[@name="foo"]/@value = '{{ param }}'`)
- Enable cells to be created dynamically in the same way
- Enable cells to be formatted within their own configuration (and remove the `format` option).

````javascript
{
    "rows": {
        {
            "cells": {
                 "one": "./expr",
                 "two": {
                     "expr": "./expr",
                     "format": ["%s%%"]
                 },
                 "three_{{ item }}": {
                     "expr": "./expr[name="{{ item }}"]",
                     "with_items": ["10", "100", "1000"]
                 }
             },
             "with_query": "//selector"
        }
        {
            "cells": {
                 "one": "//total1",
                 "one": "//total2",
            }
        }
    }
}
````

The above would add a row for each `//selector` and then a single row containing the "totals".

The "two" column would have the given `printf` formatting assigned to it and the `three` column would expand to three cells: `three_10`, `three_100` and `three_1000`.

It also removes the `headers` option (for now) and the headers are instead inferred from the column names.

This idea is influenced by the ansible configuration grammer.

### DIFF
--- a/lib/Report/Dom/PhpBenchXpath.php
+++ b/lib/Report/Dom/PhpBenchXpath.php
@@ -34,6 +34,13 @@ class PhpBenchXpath extends \DOMXpath
      */
     public function evaluate($expr, \DOMNode $context = null, $registerNodeNs = null)
     {
+        if (!is_scalar($expr)) {
+            throw new \InvalidArgumentException(sprintf(
+                __METHOD__ . ' must be passed a scalar XPath expression, got: %s',
+                print_r($expr, true)
+            ));
+        }
+
         $expr = preg_replace(
             '{php:bench\(\'([a-z]+)}', 
             'php:function(\'PhpBench\\Report\\Dom\\functions\\\$1',
--- a/lib/Report/Generator/ConsoleTableGenerator.php
+++ b/lib/Report/Generator/ConsoleTableGenerator.php
@@ -42,6 +42,7 @@ class ConsoleTableGenerator implements OutputAware, ReportGenerator
     public function configure(OptionsResolver $options)
     {
         $options->setDefaults(array(
+            'debug' => false,
             'title' => null,
             'description' => null,
             'rows' => array(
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

File: lib/Report/Dom/PhpBenchXpath.php [hunk 1]
```diff
@@ -34,6 +34,13 @@ public function __construct(\DOMDocument $dom)
      */
     public function evaluate($expr, \DOMNode $context = null, $registerNodeNs = null)
     {
+        if (!is_scalar($expr)) {
+            throw new \InvalidArgumentException(sprintf(
```
dantleech: Add test

lib/Report/Generator/ConsoleTableGenerator.php
```diff
@@ -33,6 +34,11 @@ class ConsoleTableGenerator implements OutputAware, ReportGenerator
      */
     private $formatter;
 
+    /**
+     * @var \DOMNode[]
+     */
+    private $postProcessElements;
```
dantleech: Default to array

lib/Report/Generator/ConsoleTableGenerator.php
```diff
@@ -110,13 +134,57 @@ public function generate(SuiteResult $suite, array $config)
         }
 
         $table = $this->createTable();
-        $table->setHeaders($config['headers']);
+        $table->setHeaders(array_keys($row ?: array()));
```
dantleech: Add test.

lib/Report/Generator/ConsoleTableGenerator.php
```diff
@@ -129,24 +197,82 @@ private function transformToTableDom(\DOMDocument $resultDom, \DOMDocument $tabl
...
-                } else {
-                    $cellExpr = str_replace('{selector}', $config['selector'], $cellExpr);
-                    $value = $xpath->evaluate($cellExpr, $rowEl);
-                    $this->validateXpathResult($cellExpr, $value);
-                }
 
-                $tableCellEl->setAttribute('name', $colName);
-                $tableCellEl->nodeValue = $value;
+        foreach ($config['rows'] as $rowConfig) {
+            Assert::hasOnlyKeys(array('cells', 'with_query', 'with_items'), $rowConfig, 'report config key "rows"');
```
dantleech: Will replace this with JSON schema validator in the next iteration.

lib/Report/Generator/ConsoleTableGenerator.php
```diff
@@ -284,4 +393,45 @@ private function renderTable($table)
...
+     * "context" would either be "row" or "cell".
+     *
+     * An item looks like `{{ row.item }}`.
+     *
+     * NOTE: In the future arrays could be supported, whereby the array key with dot notation
+     *       as: `{{ row.item.foobar }}`
+     *
+     * @param string $expression
+     * @param string $item
+     * @param string $context
```
dantleech: ADd return type.

PR #124, opened 2015-09-12
Better report config validation

dantleech:
Check return type from generator's `getSchema` and throw better exception message.

### DIFF
--- a/lib/Report/ReportManager.php
+++ b/lib/Report/ReportManager.php
@@ -212,15 +212,18 @@ class ReportManager
             $schema = $generator->getSchema();
             if (!is_array($schema)) {
                 throw new \InvalidArgumentException(sprintf(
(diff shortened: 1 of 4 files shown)


lib/Report/ReportManager.php
```diff
@@ -208,7 +208,19 @@ public function generateReports(OutputInterface $output, SuiteDocument $suiteDoc
...
+            if (!is_array($schema)) {
+                throw new \InvalidArgumentException(sprintf(
+                    'Generator "%s" must return the schema as an array'
+                , get_class($generator)));
+            }
+
+            $schema = json_decode(json_encode($schema));
+
+            if (empty($schema)) {
+                $schema = new \stdClass;
```
dantleech: add comment if this is really necessary

PR #125, opened 2015-09-13
Removed requirement to implement interface

dantleech:
Benchmarks are executed in separate processes, and the PHPBench library does not necessarily need to be loaded into memory, so implementing the BenchmarkInterface is an artificial requirement, and would not be found by IDEs in such cases.


examples/HashBench.php
```diff
@@ -0,0 +1,26 @@
...
+    {
+        hash('sha256', rand(0, 100000));
+    }
+
+    public function benchSha1()
+    {
+        hash('sha1', rand(0, 100000));
+    }
+}
+
```
dantleech: I guess we could add this. But this is not part of this PR. Remove extra lines.

PR #143, opened 2015-09-27
Update .styleci.yml

GrahamCampbell:
See https://blog.alt-three.com/psr4-fixer/. Today is the day the preset changes went live. Because you are using this config file rather than in-browser config, I was not able to update your config for your during the deployment.


GrahamCampbell:
You also might want to use the send fix PR functionality after merging this. :)

dantleech:
Thanks :)

PR #149, opened 2015-09-28
Added skip annotation

dantleech:
Allows methods/subjects to be skipped - useful if the method is an inherited method / abstract method.

### DIFF
--- a/examples/phpbench.json
+++ b/examples/phpbench.json
@@ -1,7 +1,6 @@
 {
     "bootstrap": "../vendor/autoload.php",
     "path": "./",
-    "progress_logger_name": "verbose",
     "outputs": {
         "console_example": {
             "renderer": "console",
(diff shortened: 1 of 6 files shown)


examples/phpbench.json
```diff
@@ -1,6 +1,7 @@
 {
     "bootstrap": "../vendor/autoload.php",
     "path": "./",
+    "progress_logger_name": "verbose",
```
dantleech: revert

lib/Benchmark/BenchmarkBuilder.php
```diff
@@ -41,6 +41,7 @@ public function build($benchmarkPath, array $subjectFilter = array(), array $gro
         );
 
         $classMeta = $this->parser->parseDoc($classInfo['comment']);
+        var_dump($classInfo['methods']);
```
dantleech: remove

dantleech:
Add doc

dantleech:
Closing in favor of #153

Issue #119, opened 2015-09-12
Overriding revs or iterations is accumulative


dantleech:
If (plural) annotations are specified at both the class and method level, should they be accumulative, or should one replace the other?

dantleech:
They should replace one another, if in the future there is a use case for specifying multiple "runs" then this can be done with an array syntax as the value of the annotation.

In the meanwhile, this can be accomplished by adding a new method with its own annotation which deleagates the subject.

dantleech:
But annotations such as `group` should be accumulative ...

dantleech:
Not since #151.

PR #151, opened 2015-10-03
Better Metadata

dantleech:
This PR aims to introduce a Metadata system, and replace the current annotation parser with an `AnnotationDriver` which uses the Doctrine annotation parser.

This will change the annotation syntax to something more standard and will solve the following issues

- Annotation accumulation (#119): Annotations will always override previous annotations. Annotations now accept arguments, e.g. `@Groups({"one", "two"})`.
- Extending classes which use common annotations (such as `@since`): The doctrine parser includes by default a large whitelist.
- Will improve the architecture and the tests.
- I guess we will be able to set an annotation namespace via. config to prevent potential clashes.

In addition it opens the way to defining metadata in other places (e.g. the config file) if required.


composer.json
```diff
@@ -16,6 +16,7 @@
         "phpbench/tabular": "^0.2",
         "seld/jsonlint": "^1.0",
         "justinrainbow/json-schema": "^1.0",
+        "doctrine/annotations": "@dev",
```
dantleech: change to stable.

lib/Benchmark/Teleflector.php
```diff
@@ -25,12 +25,18 @@ public function getClassInfo($file)
     {
         $classFqn = $this->getClassNameFromFile($file);
 
+        if (null === $classFqn) {
```
dantleech: Add test

lib/Benchmark/CollectionBuilder.php
```diff
@@ -11,21 +11,49 @@
...
-    public function __construct(BenchmarkBuilder $benchmarkBuilder, Finder $finder = null, $baseDir = null)
+    /**
+     * @var Factory
+     */
+    private $factory;
+
+    /**
+     * @param Factory $factory
+     * @param Finder $finder
+     * @param mixed $baseDir
```
dantleech: string

lib/Benchmark/Metadata/SubjectMetadata.php
```diff
@@ -0,0 +1,84 @@
...
+
+namespace PhpBench\Benchmark\Metadata;
+
+/**
+ * Metadata for benchmark subjects.
+ */
+class SubjectMetadata extends AbstractMetadata
+{
+    /**
+     * @Var string
```
dantleech: lowercase

lib/Benchmark/Remote/ReflectionHierarchy.php
```diff
@@ -0,0 +1,82 @@
...
+        foreach ($this->reflectionClasses as $reflectionClass) {
+            if (isset($reflectionClass->methods[$name])) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    public function isEmpty()
```
dantleech: doc

PR #153, opened 2015-10-06
Skip Annotation

dantleech:
This PR adds a `NotApplicable` annotation which indicates that the benchmark is not applicable to the implementation under test. Possible alternative to #149 

This feature has the added benefit that `N/A` subjects are visible in reports.


dantleech:
Currently `NotApplicable` this complicates the reports and the XML format. Maybe better to simply add the `skip` annotation in this PR.

PR #168, opened 2015-10-22
Polymorphic Executor

dantleech:
This PR makes the Executor polymorphic. The executor is responsible for executing the benchmark subject and retrieving the statistics about it.

Abstracting this makes it possible to implement different executors, such as an XDebug executor, a Blackfire executor, or whatever. Eventually even executors for different languages (JavaExecutor, PythonExecutor - but these would require alternate metadata drivers).

### DIFF
--- a/lib/Benchmark/Executor/MicrotimeExecutor.php
+++ b/lib/Benchmark/Executor/MicrotimeExecutor.php
@@ -9,9 +9,11 @@
  * file that was distributed with this source code.
  */
 
-namespace PhpBench\Benchmark;
+namespace PhpBench\Benchmark\Executor;
 
-use PhpBench\Benchmark\Metadata\SubjectMetadata;
+use PhpBench\Benchmark\ExecutorInterface;
+use PhpBench\Benchmark\Iteration;
+use PhpBench\Benchmark\IterationResult;
 use PhpBench\Benchmark\Remote\Launcher;
 
 /**
@@ -19,7 +21,7 @@ use PhpBench\Benchmark\Remote\Launcher;
  * temp. directory and then executes it. The generated script then returns the
  * time taken to execute the benchmark and the memory consumed.
  */
-class Executor
+class MicrotimeExecutor implements ExecutorInterface
 {
     /**
      * @var Launcher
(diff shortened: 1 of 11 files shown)


lib/Benchmark/Metadata/Annotations/Executor.php
```diff
@@ -0,0 +1,34 @@
...
+/*
+ * This file is part of the PHPBench package
+ *
+ * (c) Daniel Leech <daniel@dantleech.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace PhpBench\Benchmark\Metadata\Annotations;
```
dantleech: remove

lib/Benchmark/Metadata/DocParser.php
```diff
@@ -28,6 +28,7 @@ class DocParser
         'Iterations' => 'PhpBench\Benchmark\Metadata\Annotations\Iterations',
         'Revs' => 'PhpBench\Benchmark\Metadata\Annotations\Revs',
         'Skip' => 'PhpBench\Benchmark\Metadata\Annotations\Skip',
+        'Executor' => 'PhpBench\Benchmark\Metadata\Annotations\Executor',
```
dantleech: %s/d

lib/Benchmark/Metadata/Driver/AnnotationDriver.php
```diff
@@ -147,6 +147,10 @@ private function processAbstractMetadata(AbstractMetadata $metadata, $annotation
         if ($annotation instanceof Annotations\Skip) {
             $metadata->setSkip(true);
         }
+
+        if ($annotation instanceof Annotations\Executor) {
```
dantleech: d3j

lib/Benchmark/Remote/Launcher.php
```diff
@@ -38,7 +38,17 @@ public function __construct($bootstrap, $configPath)
...
+     * Launch the given template with the given parameters.
+     * 
+     * PHP ini settings can be passed to the PHP executable as a third argument.
+     *
+     * @param string $template
+     * @param array $parameters
+     * @param array $iniEntries
+     * @param string $wrapper
+     */
+    public function launch($template, array $parameters, array $iniEntries = array(), $wrapper = null)
```
dantleech: Revert this change, refactor the Launcher in a separate PR.

PR #194, opened 2015-11-12
Pass parameters as arguments

dantleech:
This PR changes the behavior of the supply of parameters to subjec, before and after methods. Previously parameters were passed as a single array, now they are expanded to the arguments of the subjects.

If the parameter set is associative, then the parameter keys will be matched to the argument names, e.g.

``` javascript
{"fooBar": "bar", "barFoo": "foo"}
```

Would require a method signature as follows:

``` php
public function benchBar($fooBar, $barFoo)
```

(the order of arguments is not important in this case).

If the parameter set is non-associative then the naming of the arguments is irrelevant.

### DIFF
--- a/lib/Benchmark/Executor/MicrotimeExecutor.php
+++ b/lib/Benchmark/Executor/MicrotimeExecutor.php
@@ -15,6 +15,7 @@ use PhpBench\Benchmark\ExecutorInterface;
 use PhpBench\Benchmark\Iteration;
 use PhpBench\Benchmark\IterationResult;
 use PhpBench\Benchmark\Remote\Launcher;
+use PhpBench\Benchmark\Metadata\SubjectMetadata;
 
 /**
  * This class generates a benchmarking script and places it in the systems
@@ -44,6 +45,11 @@ class MicrotimeExecutor implements ExecutorInterface
     public function execute(Iteration $iteration, array $options = array())
     {
         $subject = $iteration->getSubject();
+        $parameters = $this->resolveParameterOrder(
+            $subject,
+            $iteration->getParameters()
+        );
+
         $tokens = array(
             'class' => $subject->getBenchmarkMetadata()->getClass(),
             'file' => $subject->getBenchmarkMetadata()->getPath(),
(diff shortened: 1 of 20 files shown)


lib/Benchmark/Executor/MicrotimeExecutor.php
```diff
@@ -51,7 +57,8 @@ public function execute(Iteration $iteration, array $options = array())
             'revolutions' => $iteration->getRevolutions(),
             'beforeMethods' => var_export($subject->getBeforeMethods(), true),
             'afterMethods' => var_export($subject->getAfterMethods(), true),
-            'parameters' => var_export($iteration->getParameters(), true),
+            'parameters' => var_export($parameters, true),
+            'argString' => $this->getArgString(count($parameters))
```
dantleech: We construct the argument string for the template here to reduce overhead from calling `ReflectionMethod::invokeArgs`

dantleech: Add a comment about that.

lib/Benchmark/Executor/MicrotimeExecutor.php
```diff
@@ -75,4 +82,77 @@ public function getDefaultConfig()
...
+                'You have used an associative array for parameters. The argument names on the benchmarking ' . 
+                'subject "%s" must match the top level key names of the provided parameter set. Should be: "%s", current ' . 
+                'argument names: "%s"',
+                $subject->getName(),
+                implode('", "', array_keys($parameters)),
+                implode('", "', $args)
+            ));
+        }
+
+        if (count($args) && $diff = array_diff($args, array_keys($parameters))) {
```
dantleech: Document the fact that it is valid to have a zero function arity even if parameters are provided.

lib/Benchmark/Executor/MicrotimeExecutor.php
```diff
@@ -75,4 +82,77 @@ public function getDefaultConfig()
...
+                continue;
+            }
+
+            $resolved[$argIndex] = $value;
+        }
+
+        return $resolved;
+    }
+
+    private function getArgString($argCount)
```
dantleech: Add that documentation here.

lib/Benchmark/Metadata/SubjectMetadata.php
```diff
@@ -72,6 +77,17 @@ public function getParameterSets()
         return $this->parameterSets;
     }
 
+    public function getArguments()
```
dantleech: Docs

dantleech:
Add documentation.

There is infact one big problem with this approach: If the parameters are provided as parameters, then the before and after methods are coupled to the parameter providers required by any particular subject - making before and after method reuse impossible in some cases.

PR #195, opened 2015-11-12
Concurrent Iterations

dantleech:
This PR introduces the possiblity to execute iterations concurrently in order to test concurrent performance.
- Multiple concurrencies can be provided by an annotation, each will represent a new variant.
- Concurrency here is the maximum number of iterations that will be executed at the same time.
- Concurrency can be specified from the command line.
- It severely affects stability and it will take _much_ longer to arrive at a consensus if `retry_threshold` is set.


dantleech:
Could probably do this better by making the Executor return a Promise and run the benchmark process asyncronously. We could then wait until the requisite number of processes have finished before launching the next batch.

docs/writing-benchmarks.rst
```diff
@@ -253,6 +253,45 @@ Will result in the following parameter benchmark scenarios:
...
+- You specify multiple concurrency values - one variant will be created for
+  each value.
+- The number of iterations is absolute - if you have a concurrency of 5 and
+  only 3 iterations, only 3 iterations will be executed concurrently.
+- Concurrency severely affects stability, you will probably want to disable
+  the :ref:`retry_threshold`.
+
 .. _groups:
 
 Groups
```
dantleech: Update report docs to mention the new `con` column.

lib/Benchmark/Remote/Payload.php
```diff
@@ -97,18 +99,24 @@ public function launch()
...
+        $this->process->setCommandLine($wrapper . $this->phpPath . $this->getIniString() . ' ' . $this->scriptPath);
+        $this->process->start();
+
+        return $this;
+    }
 
+    public function getResult()
+    {
+        $this->process->wait();
+        unlink($this->scriptPath);
```
dantleech: This could potentially leak files -- should probably add a `__destruct` method.

PR #197, opened 2015-11-14
POC compare reports

dantleech:
This is a POC for comparing different reports.

The idea is that you run a test suite with implementation X, then on implementation Y and Z, and then compare the results.

Currently:

````
$ .phpbench compare --file=result1.xml --file=result2.xml --report=compare
````

This could be integrated directly into the `report` function however:

````bash
$ phpbench report --infile=result1.xml --infile=result2.xml --report=compare
````

The compare report would SHOULD also work on standard benchmarks (e.g. when using `run` and `report`) although its value would be degraded.

Will probably need to refactor the report format somewhat.

This work is a mess currently.

### CODE REVIEW

File: lib/Console/Command/CompareCommand.php [hunk 1]
```diff
@@ -0,0 +1,90 @@
...
+
+                if (!$suiteEl->getAttribute('name')) {
+                    $suiteEl->setAttribute('name', basename($file));
+                }
+
+                $aggregateEl->appendChild($suiteEl);
+            }
+        }
+
+        $this->reportManager->renderReports($output, $aggregateDom, $reportNames, $outputNames);
```
dantleech: This class can be removed/.

lib/Console/Command/ReportCommand.php
```diff
@@ -42,31 +43,54 @@ public function configure()
...
         $reportNames = $this->reportManager->processCliReports($reports);
-        $file = $input->getArgument('file');
+        $files = $input->getOption('file');
+
         $outputs = $input->getOption('output');
         $outputNames = $this->reportManager->processCliOutputs($outputs);
 
+        if (!$files) {
+            throw new \InvalidArgumentException(
+                'You must specify at least one result --file (generated by the run command\'s `--dump-file=` option)'
```
dantleech: Test?

lib/Extension/CoreExtension.php
```diff
@@ -191,6 +192,12 @@ private function registerCommands(Container $container)
                 $container->get('report.manager')
             );
         }, array('console.command' => array()));
+
+        $container->register('console.command.compare', function (Container $container) {
+            return new CompareCommand(
```
dantleech: REmove

lib/Report/Generator/TabularGenerator.php
```diff
@@ -59,8 +59,14 @@ public function getSchema()
                 'pretty_params' => array(
                     'type' => 'boolean',
                 ),
-                'aggregate' => array(
-                    'type' => 'boolean',
+                'mode' => array(
```
dantleech: Document

dantleech: Is compare required?

lib/Benchmark/Runner.php
```diff
@@ -162,13 +162,21 @@ public function setRetryThreshold($retryThreshold)
     /**
      * Run all benchmarks (or all applicable benchmarks) in the given path.
      *
-     * @param string
+     * The $name argument will set the "name" attribute on the "suite" element.
+     *
+     * @param string $name
```
dantleech: fix doc

dantleech:
Remove context from aggregate/defautl reports until we can inteligently hide empty columns.

PR #204, opened 2015-11-16
Delimited output

dantleech:
This PR adds the "delimited" renderer, allowing the tables to be rendered as a delimited list.

In addition this PR:

- Adds a `plain` report -- the same as default but without the footer
- Modifies the behavior of the `quiet` option to NOT suppress report output.


lib/Report/Generator/TabularGenerator.php
```diff
@@ -83,6 +83,12 @@ public function getSchema()
                         array('type' => 'null'),
                     ),
                 ),
+                'formatting' => array(
+                    'type' => 'boolean',
+                ),
+                'footer' => array(
```
dantleech: Rename to body only

PR #234, opened 2015-12-14
Runner context

dantleech:
Introduces a runner context class for configuring the runner.

### DIFF
--- a/lib/Benchmark/RunnerContext.php
+++ b/lib/Benchmark/RunnerContext.php
@@ -1,0 +1,224 @@
+<?php
+
+/*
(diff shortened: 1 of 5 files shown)


lib/Benchmark/RunnerContext.php
```diff
@@ -0,0 +1,195 @@
...
+     *
+     * A value of NULL will disable retry.
+     *
+     * @return float
+     */
+    public function getRetryThreshold($default)
+    {
+        return $this->options['retry_threshold'] ?: $default;
+    }
+}
```
dantleech: Document the role of `$default` in this class.

PR #243, opened 2015-12-21
Benchmark mode

### LINKED ISSUE #237: Throughput mode
Sometimes it may be more meaningful to display results as operations per second as opposed to showing the averge time per operation. JMH does this with the [@BenchmarkMode](http://java-performance.info/jmh/) annotation.

This should change the output on both the reports and the progress loggers to show ops/s (or rps, or even hertz).

dantleech:
Specify if output should be displaed as "ops per second" (throughput) or "time per op".

Introduces the new `@OutputMode("throughput")` annotation and the `output` mode configuration option.

Would fix #237


lib/Benchmark/Metadata/AbstractMetadata.php
```diff
@@ -67,6 +67,11 @@
     private $outputTimeUnit = null;
 
     /**
+     * @var string
+     */
+    private $outputMode = 'time';
```
dantleech: Use constant

tests/Unit/Util/TimeUnitTest.php
```diff
@@ -70,12 +70,54 @@ public function provideConvert()
     }
 
     /**
+     * It should convertTo one time unit to another.
+     *
+     * @dataProvider provideConvertHertz
+     */
+    public function testConvertHertz($time, $unit, $destUnit, $expectedHertz)
```
dantleech: Not hertz

dantleech:
Consider combining the TimeFormatter into the TimeUnit class, maybe renaming that.

dantleech:
Integrated into the main application, now need to integrate the functionality into the reports.

lib/Progress/Logger/PhpBenchLogger.php
```diff
@@ -53,18 +53,17 @@ public function endSuite(SuiteDocument $suiteDocument)
...
         ));
         $this->output->writeln(sprintf(
-            '⅀T: %s%s μSD/r %s%s μRSD/r: %s%%',
-            number_format($this->timeUnit->value($suiteDocument->getTotalTime()), 3),
-            $this->timeUnit->getDestSuffix(),
-            number_format($this->timeUnit->value($suiteDocument->getMeanStDev()), 3),
-            $this->timeUnit->getDestSuffix(),
+            '⅀T: %s μSD/r %s μRSD/r: %s%%',
+            $this->timeUnit->format($suiteDocument->getTotalTime(), null, TimeUnit::MODE_TIME, true),
+            $this->timeUnit->format($suiteDocument->getMeanStDev(), null, TimeUnit::MODE_TIME, true),
```
dantleech: Last argument is useless.

PR #295, opened 2016-01-14
escape script path when calling command

Remo:
I had escape this in order to get it working on my Windows box. My temp dir is in the user directory and there's a blank in the path. This fixes the problem and should work on *nix too in case someone has a blank in their temp path, probably never happens there though.

### DIFF
--- a/lib/Benchmark/Remote/Payload.php
+++ b/lib/Benchmark/Remote/Payload.php
@@ -109,7 +109,7 @@ class Payload
             $wrapper = $this->wrapper . ' ';
         }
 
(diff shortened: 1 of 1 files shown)


dantleech:
Looks sensible :) Would be awesome if you could add a test for this in: https://github.com/phpbench/phpbench/blob/master/tests/System/RunTest.php

dantleech:
https://github.com/phpbench/phpbench/issues/296

Remo:
thanks for the quick merge! I'll see what I can do about the test, will create some benchmarks for https://github.com/punic/punic first ;-)

PR #297, opened 2016-01-14
add composer.lock to repository

Remo:
`composer.lock` should be included in the repo. Otherwise you might be running tests on travis with different dependencies. By including `composer.lock` you can make sure you test what you've been developing with.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
-composer.lock
 vendor
 docs/build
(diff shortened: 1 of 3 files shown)


dantleech:
PHPBench is essentially a library, we don't commit `composer.lock` to libraries. The library should work with the range of dependencies specified in the composer.json file, it is not tied to any particular version.

Travis checks both with the "latest" library versions and the lowest boundry.

dantleech:
See the note at the bottom of: https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file. But thanks for the PR :)

Remo:
okay, thanks (:

PR #298, opened 2016-01-14
fix unit tests when using windows

Remo:
It's not complete, still some broken tests. Mostly related to console and processes, one example below. I'll try to continue when I'm home.

```
2) PhpBench\Tests\System\RunTest::testCommandWithReportConfigurationUnknown
Failed asserting that '

  [Symfony\Component\Console\Exception\RuntimeException]
  Too many arguments.
```

### DIFF
--- a/bin/phpbench.bat
+++ b/bin/phpbench.bat
@@ -1,0 +1,2 @@
+@echo off
+php.exe "%~dp0\phpbench.php" %*
--- a/tests/System/SystemTestCase.php
+++ b/tests/System/SystemTestCase.php
@@ -25,7 +25,7 @@ class SystemTestCase extends \PHPUnit_Framework_TestCase
     public function createResult($benchmark = 'benchmarks/set4')
     {
         $process = $this->phpbench(
-            'run ' . $benchmark . ' --executor=debug --dump-file=' . $this->fname
+            'run ' . $benchmark . ' --executor=debug --dump-file=' . escapeshellarg($this->fname)
         );
 
         if ($process->getExitCode() !== 0) {
@@ -49,6 +49,7 @@ class SystemTestCase extends \PHPUnit_Framework_TestCase
     {
         chdir($this->getWorkingDir($workingDir));
         $bin = __DIR__ . '/../../bin/phpbench --verbose';
+        $bin = str_replace('/', DIRECTORY_SEPARATOR, $bin);
         $process = new Process($bin . ' ' . $command);
         $process->run();
 


bin/phpbench.bat
```diff
@@ -0,0 +1,2 @@
+@echo off
+php.exe "%~dp0\phpbench.php" %*
```
dantleech: Is this file necessary? I notice that well established libraries such as phpspec do not have this file. https://github.com/phpspec/phpspec/tree/master/bin

tests/System/SystemTestCase.php
```diff
@@ -49,6 +49,7 @@ public function phpbench($command, $workingDir = '.')
     {
         chdir($this->getWorkingDir($workingDir));
         $bin = __DIR__ . '/../../bin/phpbench --verbose';
+        $bin = str_replace('/', DIRECTORY_SEPARATOR, $bin);
```
dantleech: Hmm, I thought that `/` worked in Windows these days, there are many instances of this. Which version of PHP are you using?

bin/phpbench.bat
Remo: Kind of. I can do this:
```
C:\Work\phpbench>bin\phpbench
phpbench version 0.10.0-dev
```

But when starting a process using `new Process` it doesn't work. Removing that bat file changes this:
`Tests: 245, Assertions: 461, Errors: 17, Failures: 31.`
to this:
`Tests: 245, Assertions: 504, Errors: 5, Failures: 15.`

tests/System/SystemTestCase.php
Remo:
```
C:\Work\phpbench>php -v
PHP 5.6.12 (cli) (built: Aug  6 2015 12:06:17)
Copyright (c) 1997-2015 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2015 Zend Technologies
    with Xdebug v2.3.3, Copyright (c) 2002-2015, by Derick Rethans
```

Never heard that you can use `/` in the console. I'm on Windows 10:
```
C:\Work\phpbench>bin/phpbench
'bin' is not recognized as an internal or external command,
operable program or batch file.

C:\Work\phpbench>bin\phpbench
phpbench version 0.10.0-dev
...
```

Remo: I just fixed this one occurence because I wanted to know how/if you'd want to take care of it. I've seen a few more instances. Most unit tests are working though, it's not that bad.

bin/phpbench.bat
dantleech: Strange, I really need to get a Windows VM setup locally :)

Remo: Sorry, the above is wrong. This doesn't work `C:\Work\phpbench>bin\phpbench`.
Too many open projects /-:
It works because composer creates a bat file when installing it as a library. https://getcomposer.org/doc/articles/vendor-binaries.md#what-about-windows-and-bat-files-

However, when I run `composer update` it doesn't seem to create that file.

tests/System/SystemTestCase.php
Remo: You probably have something slightly different in mind. Functions like `fopen` can handle `/`, but here we are starting a process, nothing related to PHP, only about processes / parameter.

dantleech: Ah ok, yes your right. I would probably go for `implode(DIRECTORY_SEPARATOR, array('path', 'to', 'something'))` in that case.

bin/phpbench.bat
dantleech: Maybe try removing the vendor directory, removing composer.lock and
doing a `composer install`, perhaps ?

Remo: I'm afraid, no luck.

WouterJ: Just use `php bin/phpbench` and things work perfectly :)

Remo: @WouterJ that only works if you use it as a library, composer doesn't create those for standalone projects

PR #308, opened 2016-01-28
Model metadata

dantleech:
This PR separates the roles of Metadata and Model.

The Metadata has all the information needed to *construct* the model and run the benchmarks. The Model is just the representation of the data and the logic which goes with it (e.g. calculating rejects, statistics, etc.)

### DIFF
--- a/lib/Benchmark/Executor/BaseExecutor.php
+++ b/lib/Benchmark/Executor/BaseExecutor.php
@@ -12,9 +12,10 @@
 namespace PhpBench\Benchmark\Executor;
 
 use PhpBench\Benchmark\ExecutorInterface;
+use PhpBench\Benchmark\Metadata\BenchmarkMetadata;
+use PhpBench\Benchmark\Metadata\SubjectMetadata;
 use PhpBench\Benchmark\Remote\Launcher;
 use PhpBench\Benchmark\Remote\Payload;
-use PhpBench\Model\Benchmark;
 use PhpBench\Model\Iteration;
 use PhpBench\Model\IterationResult;
 use PhpBench\Registry\Config;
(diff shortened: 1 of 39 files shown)


lib/Benchmark/Executor/BaseExecutor.php
```diff
@@ -51,18 +52,17 @@ public function __construct(Launcher $launcher)
...
-            'afterMethods' => var_export($subject->getAfterMethods(), true),
-            'parameters' => var_export($iteration->getParameters()->getArrayCopy(), true),
-            'warmup' => $iteration->getWarmup() ?: 0,
+            'class' => $subjectMetadata->getBenchmark()->getClass(),
+            'file' => $subjectMetadata->getBenchmark()->getPath(),
+            'subject' => $subjectMetadata->getName(),
+            'revolutions' => $subjectMetadata->getRevs(),
+            'beforeMethods' => var_export($subjectMetadata->getBeforeMethods(), true),
+            'afterMethods' => var_export($subjectMetadata->getAfterMethods(), true),
+            'parameters' => var_export($iteration->getVariant()->getParameterSet()->getArrayCopy(), true),
```
dantleech: Could just as well pass the `Variant` instead of the Iteration?

lib/Model/Suite.php
```diff
@@ -58,6 +54,19 @@ public function getBenchmarks()
         return $this->benchmarks;
     }
 
+    /**
+     * Create and add a benchmark.
+     *
+     * @return Benchmark
+     */
```
dantleech: `@param` ..

lib/Model/Suite.php
```diff
@@ -145,4 +144,14 @@ public function getErrorStacks()
 
         return $errorStacks;
     }
+
+    public function setEnvInformations(array $envInformations)
```
dantleech: dc.

dantleech:
Timeunit doesn't seem to be persisted to XML

PR #312, opened 2016-01-29
Encode decode

dantleech:
Reporting now accepts `SuiteCollection` model as argument instead of a DOMDocument.

- The core is not knowing about the XML document (in principly - we still dump to XML from the run command for now).
- Introduced an XML decoder - used by the report framework when reading XML files.


lib/Math/Statistics.php
```diff
@@ -102,7 +102,7 @@ public static function mean($values)
      *
      * @return float[]
      */
-    public static function kdeMode(array $population, $space = 512, $bandwidth = null)
+    public static function kdeMode(array $population, $space = 10, $bandwidth = null)
```
dantleech: Revert

PR #315, opened 2016-01-30
Non-strict "JSON" support.

dantleech:
- New JsonDecoder.
- Allows unquoted keys.
- Allows non encapsulated strings.
- Parses using JsonParse.
- Applies to all things that can be configured from the CLI with JSON.

From:

```bash
$ phpbench run --report='{"extends": "aggregate", "exclude": ["benchmark", "params"]}'
```

To:

```bash
$ phpbench run --report='extends: "aggregate", exclude: ["benchmark", "params"]'
$ phpbench run --executor='extends: "debug", times: [1, 2, 3, 4], spread: [0, 1, 2]'
```

### DIFF
--- a/lib/Json/JsonDecoder.php
+++ b/lib/Json/JsonDecoder.php
@@ -1,0 +1,77 @@
+<?php
+
+/*
+ * This file is part of the PHPBench package
+ *
+ * (c) Daniel Leech <daniel@dantleech.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace PhpBench\Json;
+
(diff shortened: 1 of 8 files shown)


lib/Json/JsonDecoder.php
```diff
@@ -0,0 +1,77 @@
...
+        return json_decode($jsonString, true);
+    }
+
+    /**
+     * Taken from: https://gist.github.com/larruda/967110d74d98c1cd4ee1.
+     */
+    private function normalize($jsonString)
+    {
+        if (!is_string($jsonString)) {
+            throw new \Exception(sprintf(
```
dantleech: Test

PR #340, opened 2016-02-28
Added baseline env provider

dantleech:
<strike>This PR is attempting to establish a good strategy for baseline measurements. It currently implements two new features, although one of them will probably be dropped in favor of the other.</strike>. This PR is now only abount the env provider. The other strategy is a [https://github.com/phpbench/phpbench/issues/372](separate feature).

The idea is to provide a way to better understand results taken from different machines, or anomalous results taken on the same machine.

## Environment Provider

Adds an Envrionment Provider which runs some very simple "benchmarks" (md5 10000 times, nothing 10000 times) and provides the time in microseconds.

All information made available through Environment providers are available in reports.

This provides a "better-than-nothing" indication of the performance of the machine the benchmarks are running on.

```
(standard report, selecting the `baseline_nothing` col)
+------------------+------------+------------+
| baseline_nothing | other cols | mean       |
+------------------+------------+------------+
| 0.000045ms       | ...        | 0.019320ms |
+------------------+------------+------------+

(env report)
Suite # 2016-02-28 09:32:05
+--------------+---------+------------------------------------------+
| provider     | key     | value                                    |
+--------------+---------+------------------------------------------+
| uname        | os      | Linux                                    |
| ...          | ...     | ...                                      |
| vcs          | version | fe14d5c1ee747a52cb1074d11bd6617dd31e38aa |
| baseline     | md5     | 0.29157161712646                         |
| baseline     | nothing | 0.044848918914795                        |
+--------------+---------+------------------------------------------+
```

## Per-iteration baseline

See: https://github.com/phpbench/phpbench/issues/372<S-Del>

### DIFF
--- a/lib/Benchmark/Baseline/Baselines.php
+++ b/lib/Benchmark/Baseline/Baselines.php
@@ -1,0 +1,65 @@
+<?php
+
+/*
+ * This file is part of the PHPBench package
+ *
+ * (c) Daniel Leech <daniel@dantleech.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace PhpBench\Benchmark\Baseline;
+
+/**
+ * Class providing some static methods which are used
+ * to provide base line measurements.
+ *
+ * @see \PhpBench\Benchmark\BaselineManager
+ */
+class Baselines
+{
+    /**
+     * Do nothing.
+     */
+    public static function nothing($revs)
+    {
+        for ($i = 0; $i < $revs; $i++) {
+        }
+    }
+
+    /**
+     * Calculate an md5 hash.
+     */
+    public static function md5($revs)
+    {
+        for ($i = 0; $i < $revs; $i++) {
+            md5('lorem ipusm');
+        }
+    }
+
+    /**
+     * Open a file, write a string to it $revs times, then
+     * read each line back.
+     */
+    public static function fwriteFread($revs)
+    {
+        $tempName = tempnam(sys_get_temp_dir(), 'phpbench_baseline');
+        $handle = fopen($tempName, 'w');
+
+        for ($i = 0; $i < $revs; $i++) {
+            fwrite($handle, 'lorum ipsum');
+        }
+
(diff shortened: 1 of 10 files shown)


lib/Environment/Provider/Baseline.php
```diff
@@ -0,0 +1,47 @@
...
+    {
+        return true;
+    }
+
+    public function getInformation()
+    {
+        return new Information(
+            'baseline',
+            array(
+                'md5' => $this->measure(function () { md5('hello_world'); }),
```
Ocramius: Is an operation causing basic i/o worth reporting?

dantleech: Yeah, at least with the envrionment provider it would make sense.

dantleech:
We can scale the baseline / measurement and offset the offset, in the above example that would produce a much more stable line.

If we can also store a disk I/O metric for each iteration. Some benchmarks may be affected more by disk I/O than CPU. We should be able to determine *how much* by fitting the measurements line over the disk and CPU line and see which affected it more and apply the "offset" of each depending on the amount it seems to correlate with the measurements.

lib/Benchmark/Baseline/Baselines.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+namespace PhpBench\Benchmark\Baseline;
+
+class Baselines
```
dantleech: docs

lib/Benchmark/BaselineManager.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+namespace PhpBench\Benchmark;
+
+class BaselineManager
```
dantleech: doc

dantleech:
Updated, I have dropped the per-iteration thing and made that a separate issue, also added a simple file I/O test:

```
// env report
// ..
| baseline     | nothing | 0.14686584472656                                   |
| baseline     | md5     | 1.2390613555908                                    |
| baseline     | file_rw | 2.1030902862549                                    |
+--------------+---------+----------------------------------------------------+
```

PR #347, opened 2016-03-02
[WIP] XML storage

dantleech:
An XML storage layer for persisting and querying benchmark results.

- [x] Store in XML
- [x] Implement history
- [ ] Query ...

This PR tackles the problem of storing XML documents as follows:

- It creates a single XML file per-suite with a naming schema `<run ID><date><vcs_branch>.xml`
- Queries will be converted to XPath queries and then executed against a PHPBench XML document which contains all of the suites relevant to the query.

The main issues here are:

1. The need to store and *order* an unbounded number of XML documents. (note that if GIT storage is implemented around this driver, the number will only continue to rise).
2. The need to reduce the number of XML documents prior to querying.

(1.) is required so that benchmarks can be listed in order using the `history` command, and it means reading and then ordering all of the filenames (which is perhaps not so bad).

(2.) requires the query to be analyzed to return the limiting criteria (run ID should be in 1,2,3 AND date > 2015 AND date < 2018 OR date IN 2015-01-01, 2016-02-01 etc) - we can then filter the filenames of the XML documents we should load.

### DIFF
--- a/lib/Extension/CoreExtension.php
+++ b/lib/Extension/CoreExtension.php
@@ -445,6 +445,20 @@ class CoreExtension implements ExtensionInterface
         $container->register('storage.driver_factory', function (Container $container) {
             return new Storage\DriverFactory($container, $container->getParameter('storage'));
         });
+
+        $container->register('storage.driver.xml.persister', function (Container $container) {
+            return new Storage\Driver\Xml\Persister(
+                $container->get('serializer.encoder.xml'),
+                'foo'
+            );
+        });
+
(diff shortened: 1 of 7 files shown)


dantleech:
I wonder if this wouldn't be easier as an archive function. i.e archive from a storage driver to XML.

dantleech:
Should probably close this and make it read only, then focus on archiving.

dantleech:
Closed in favor of https://github.com/phpbench/phpbench/pull/376

PR #348, opened 2016-03-02
Archive and restore

dantleech:
This feature allows the contents of the storage driver to be dumped to PHPBench XML files and then later restored.

XML is the default archiving implementation, but others may be implemented and configured - notably a GIT archiver (which will use the XML archiver).

This will permit:

- Serialization of benchmarking results in plain text (i.e. possiblity to store in GIT).
- Changing storage drivers.

```bash
$ phpbench archive
$ phpbench archive --restore
```

## TODO

- [x] Strore unique hash as run ID (based on GIT hash)
- [x] Store / restore
- [x] Skip existing entries (on both operations)
- [x] Archive "registry" - eventually be able to choose GIT archiving strategy.

## Next steps:

- Optimize Sqlite driver insertion (batch insert for suites / benchmarks / subjects / variants)
- "Delete" query on storage layer - need ability to be able to remove results from storage (and so be able to have some control over the archive by regeneration).
- GIT archive

### DIFF
--- a/docs/storage.rst
+++ b/docs/storage.rst
@@ -78,9 +78,6 @@ existing :doc:`reports`:
 
     $ phpbench report --query='run: 239' --report=aggregate
 
-Query Language
---------------
-
 PHPBench uses a query language very similar to that of MongoDB. A simple
 example:
 
@@ -172,7 +169,7 @@ queries::
     benchmark: { $regex: "Foo.*Bench" }
 
 Fields
-------
+~~~~~~
 
 The following fields are currently available for querying:
 
@@ -190,3 +187,39 @@ be enclosed in square brackets as follows::
     param[nb_elements]: 10
 
     param[points]: { $gt: 50 }
+
+Archiving
+---------
+
+Archiving provides a way to export and reimport data from and to the
+configured storage. This allows you to:
+
+- Backup your results (for example to a GIT repository).
+- Migrate to other storage drivers.
+
+By default PHPBench is configured to use an ``XML`` archiver, which will dump
+results to a directory in the current working directory, ``_archive``.
+
+.. note::
+
+    In the future it will be possible to select GIT as an archiving
+    mechanism, so that your benchmarks are always stored in your codes GIT
+    repository.
+
+To archive::
+
+    $ phpbench archive
+
+To restore::
+
+    $ phpbench archive --restore 
+
+Both operations are idempotent - they will skip any existing records.
+
+You may configure a different archiver in the configuration:
+
+.. code-block:: javascript
+
+    {
+        "archiver": "xml"
+    }
--- a/lib/Console/Command/ArchiveCommand.php
+++ b/lib/Console/Command/ArchiveCommand.php
@@ -1,0 +1,69 @@
+<?php
+
+/*
+ * This file is part of the PHPBench package
+ *
+ * (c) Daniel Leech <daniel@dantleech.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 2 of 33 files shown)


dantleech:
@ocramius what do you think to this approach? 

This allows us to serialize from Storage to XML, and later restore. The problem with an XML *storage* is the complexity of querying. 

We could still implement a *write only* XML driver, user can then later restore to a readable/queryable storage driver and generate reports.

dantleech:
Also thinking I should just adopt Doctrine DBAL and generic database support.

lib/Model/Suite.php
```diff
@@ -185,4 +185,26 @@ public function getIdentifier()
...
+     *
+     * Based on the GIT hashing detailed here:
+     *     http://alblue.bandlem.com/2011/08/git-tip-of-week-objects.html
+     */
+    public function generateIdentifier()
+    {
+        $serialized = serialize([
+            microtime(),
+            $this->envInformations,
+            $this->configPath
```
dantleech: only need to serialize `envInformations` here, others can go into `sha1` below.

lib/Serializer/XmlEncoder.php
```diff
@@ -43,6 +43,7 @@ public function encode(SuiteCollection $suiteCollection)
             $suiteEl->setAttribute('context', $suite->getContextName());
             $suiteEl->setAttribute('date', $suite->getDate()->format('Y-m-d H:i:s'));
             $suiteEl->setAttribute('config-path', $suite->getConfigPath());
+            $suiteEl->setAttribute('id', $suite->getIdentifier());
```
dantleech: sute id or run id?

lib/Console/Command/ArchiveCommand.php
```diff
@@ -0,0 +1,55 @@
...
+
+use PhpBench\Storage\DriverFactory;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use PhpBench\Storage\Archiver;
+use PhpBench\Registry\Registry;
+
+class ArchiveCommand extends Command
```
dantleech: Document and update command description and help

tests/System/output/markdown.md
```diff
@@ -1,7 +1,7 @@
 PHPBench Benchmark Results
 ==========================
 
-### suite: 
+### suite: %run_id%
```
dantleech: It would seem then that the other outputs do not container the suite hash.

dantleech:
Need to fix the history display, as it does not look good with the longer hash identifier.

docs/storage.rst
```diff
@@ -190,3 +187,43 @@ be enclosed in square brackets as follows::
...
+
+To archive::
+
+    $ phpbench archive
+
+To restore::
+
+    $ phpbench archive --restore 
+
+In both operations are idempotent - they will skip any existing records.
```
dantleech: s/In both/Both

docs/storage.rst
```diff
@@ -190,3 +187,43 @@ be enclosed in square brackets as follows::
...
+are stored in a binary format which does not work well in some storage systems
+(e.g. GIT).
+
+Additionally you may wish to change your local storage system.
+
+Archiving is the process of exporting all of the data from a Storage driver to
+plain text files and the reverse operation: restoration.
+
+By default PHPBench is configured to use an ``XML`` archiver, which will dump
+results to a directory in the current working directory, ``_archive``.
```
dantleech: Revise this wording, it is strange

PR #349, opened 2016-03-02
Wraps template contents into immediately invoked functions

lmuzinic:
There was a variable name conflict between phpbench's $file in Remote/templates and our project's $file in config loader.

### DIFF
--- a/lib/Benchmark/Executor/template/benchmark_static_methods.template
+++ b/lib/Benchmark/Executor/template/benchmark_static_methods.template
@@ -1,20 +1,22 @@
 <?php
 
-$class = '{{ class }}';
-$file = '{{ file }}';
-$methods = {{ methods }};
-$bootstrap = '{{ bootstrap }}';
+call_user_func(function() {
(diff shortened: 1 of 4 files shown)


dantleech:
Hmm, does this really work?

I was thinking more:

```php
<?php
// test.php

$foo = 'bar';
```

```php
$foo = 'hello';

function () {
    require_once('test.php');
};

echo $foo; // "hello"
```

dantleech:
Would be great to add a regression test also - although not sure the best way to do that, as this issue is directly related to the script "template" files.

Perhaps for now it will be ok just to add a simple regression test [here](https://github.com/phpbench/phpbench/blob/master/tests/System/RunTest.php#L225) - and provide a bootstrap file which assigns `$file` to something else.

lmuzinic:
OK, thanks for feedback. I'll refactor and add tests.

dantleech:
Closing in favor of: https://github.com/phpbench/phpbench/pull/393

PR #377, opened 2016-03-26
Support for deleting runs

dantleech:
This PR adds support for deleting runs from the storage.

- New `delete` command, accepts either queries or UUIDs
- Delete implmented for both XML and DBAL.
- Generic SuiteCollection handler also now accepts UUIDs - therefore the report command can be used with the XML storage driver.


dantleech:
Need to add functional tests for DBAL delete, etc.

PR #411, opened 2016-05-27
Phpunit/Generic class support

dantleech:
There is no important reason that PHPUnit classes cannot be used as benchmarks.

This PR will allow PHPBench annotations to be imported into PHPUnit tests, 

```php
    /**
     * It should measure the mean time taken to execute a callable.
     *
     * @\PhpBench\Benchmark\Metadata\Annotations\Subject()
     * @\PhpBench\Benchmark\Metadata\Annotations\Iterations(100)
     */
    public function testCallable()
    {
        static::$callCount = 0;
        $this->manager->addBaselineCallable('foo', __CLASS__ . '::baselineExample');
        $this->manager->benchmark('foo', 100);
        $this->assertEquals(100, static::$callCount);
    }
```


lib/Benchmark/Metadata/Annotations/Benchmark.php
```diff
@@ -0,0 +1,20 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace PhpBench\Benchmark\Metadata\Annotations;
+
+/**
+ * @Annotation
+ * @Taget({"CLASS"})
+ */
+class Benchmark
```
dantleech: Not required.

lib/Benchmark/Metadata/BenchmarkMetadata.php
```diff
@@ -42,6 +42,11 @@ class BenchmarkMetadata
     private $afterClassMethods = [];
 
     /**
+     * @var bool
+     */
+    private $isBenchmark = false;
```
dantleech: Not required.

lib/Benchmark/Metadata/DocParser.php
```diff
@@ -13,13 +13,16 @@
 
 use Doctrine\Common\Annotations\AnnotationException;
 use Doctrine\Common\Annotations\DocParser as DoctrineDocParser;
+use Doctrine\Common\Annotations\PhpParser;
 
 /**
  * PHPBench doc parser, uses the Doctrine DocParser.
  */
 class DocParser
```
dantleech: Probably need to refactor here and either use the Doctrine `AnnotationReader` (there was a reason I didn't use that) or adopt the same pattern.

dantleech: As we need to import the `use` statements from the PHP file.

lib/Benchmark/Metadata/AnnotationReader.php
```diff
@@ -75,6 +74,9 @@ class DocParser
         'var' => true, 'version' => true,
         // PHPUnit tags
         'codeCoverageIgnore' => true, 'codeCoverageIgnoreStart' => true, 'codeCoverageIgnoreEnd' => true,
+        'expectedException' => true, 'expectedExceptionMessage' => true,
+        'dataProvider' => true,
+        // TODO: Add all PHPUnit annotations here ...
```
dantleech: Add __all__ the PHPUnit annotations.

PR #413, opened 2016-05-30
Allow configuration of launcher.

dantleech:
- Specify PHP binary
- Specify PHP ini settings
- Specify PHP wrapper

CLI:

```bash
$ phpbench run --php-binary=php7.0 --php-wrapper="blackfire run" --php-ini='memory_limit: 500, foobar: "foo"'
```

Or in configuration:

```yaml
{
    "php_binary": "hhvm",
    "php_wrapper": "tideways",
    "php_ini": {
        "memory_limit": 10
    }
}
```

### DIFF
--- a/lib/Benchmark/Remote/Launcher.php
+++ b/lib/Benchmark/Remote/Launcher.php
@@ -11,6 +11,9 @@
 
 namespace PhpBench\Benchmark\Remote;
 
+use Symfony\Component\Process\ExecutableFinder;
+use Symfony\Component\Process\PhpExecutableFinder;
+
 /**
  * Build and execute tokenized scripts in separate processes.
  * The scripts should return a JSON encoded string.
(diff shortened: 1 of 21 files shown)


dantleech:
With this PR the environment that is collected and the one the benchmarks are executed in could be different -- we need to collect the PHP runtime information from the launcher, not the current process.

lib/Benchmark/Remote/Launcher.php
```diff
@@ -43,6 +77,34 @@ public function payload($template, array $tokens)
...
+            } else {
+                $phpBinary = $this->finder->find($this->phpBinary);
+            }
+
+            if (null === $phpBinary) {
+                throw new \InvalidArgumentException(sprintf(
+                    'Could not find PHP binary "%s"', $this->phpBinary
+                ));
+            }
+        } else {
```
dantleech: remove the `else`

lib/PhpBench.php
```diff
@@ -142,6 +158,11 @@ private static function loadConfig()
             break;
         }
 
+        $config = array_merge(
+            $config,
+            $configOverride
+        );
+
         if ($bootstrapOverride) {
             $config['bootstrap'] = self::getBootstrapPath(getcwd(), $bootstrapOverride);
```
dantleech: can be part of `$configOverride`

lib/Benchmark/Remote/Launcher.php
```diff
@@ -43,6 +77,45 @@ public function payload($template, array $tokens)
...
+        // resolve to PHP_BINARY)
+        if (!$this->phpBinary) {
+            $finder = new PhpExecutableFinder();
+
+            return $finder->find();
+        }
+
+        // if the php binary is absolute, fine.
+        if (substr($this->phpBinary, 0, 1) === '/') {
+            return $this->phpBinary;
```
dantleech: Check for existence

PR #421, opened 2016-06-16
XDebug trace support

dantleech:
This PR will add a new executor - `xdebug_trace` which will generate extra `xdebug` results including the sum of (according to XDebug) memory and time, more usefully it will add the number of function calls - usefully these metrics are automatically filtered to exclude all calls outside of the benchmark method itse.f

These metrics will be available to standard reports. In addition there will be a command to show a function trace.


extensions/xdebug/lib/Command/TraceCommand.php
```diff
@@ -0,0 +1,108 @@
...
+     */
+    public function execute(InputInterface $input, OutputInterface $output)
+    {
+        $this->output = $output;
+        $outputDir = $input->getOption('outdir') ?: $this->outputDir;
+        $dump = $input->getOption('dump');
+
+        if (!$this->filesystem->exists($outputDir)) {
+            $output->writeln(sprintf(
+                '<comment>// creating non-existing directory %s</comment>',
```
dantleech: Duplicated with the profile commandg

extensions/xdebug/lib/Executor/TraceExecutor.php
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefaultConfig()
+    {
+        return [
+            'callback' => function () {
+            },
+            'output_dir' => 'profile',
```
dantleech: xdebug

extensions/xdebug/lib/Executor/template/xdebug.template
```diff
@@ -0,0 +1,45 @@
+<?php
+
```
dantleech: add comment - this is the same as the standard executor template but doesn't return anything

extensions/xdebug/tests/System/TraceCommandTest.php
```diff
@@ -0,0 +1,36 @@
...
+
+class TraceCommandTest extends XDebugTestCase
+{
+    /**
+     * It should run when given a path.
+     * It should show the default (simple) report.
+     */
+    public function testCommand()
+    {
+        $process = $this->phpbench('xdebug:trace benchmarks/set1/BenchmarkBench.php --filter=benchDoNothing');
```
dantleech: Test all the options

PR #428, opened 2016-07-02
Lazy build

dantleech:
- Do not eagerly instantaite everything.
- Need to make ConfigurableRegistry lazy too ...


lib/Registry/ConfigurableRegistry.php
```diff
@@ -133,12 +124,25 @@ private function resolveConfig(array $config)
...
+                '%s configuration must EITHER indicate its target %s service with the "%s" key or extend an existing configuration with the "extends" key.',
+                $this->serviceType,
+                $this->serviceType,
+                $this->serviceType
+            ));
+        }
+
+        $service = $this->getService($config[$this->serviceType]);
+
+        // eagerly validate
```
dantleech: no longer eager validation

PR #429, opened 2016-07-02
Using better reflection

dantleech:
This PR uses BetterReflection instead of PHPBenchs homegrown reflection.

This can have a significant performance impact when the target is using composer - in which case we do not need to spawn new PHP processes to get the reflection data.

If projects are not using composer then we can either source a single file, or revert to spawning a new process using a new "process locator".

Initially this is 2x as fast on Linux, on Windows / HHVM it would be faster still.


lib/Reflection/ComposerFileReflector.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        $this->reflector = new ClassReflector(new ComposerSourceLocator($classLoader));
+    }
+
+    public function reflectFile($file)
+    {
+        $class = $this->getClassNameFromFile($file);
+
+        if (null === $reflection = $this->reflector->reflect($class)) {
+            $locator = new SingleFileSourceLocator($file);
```
dantleech: Need to test and perhaps review this scenario.

lib/Reflection/Locator/RemoteSourceLocator.php
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+namespace PhpBench\Reflection\Locator;
+
+use BetterReflection\Identifier\Identifier;
+use BetterReflection\SourceLocator\Located\LocatedSource;
+use BetterReflection\SourceLocator\Type\AbstractSourceLocator;
+use PhpBench\Benchmark\Remote\Launcher;
+
+class RemoteSourceLocator extends AbstractSourceLocator
```
dantleech: Need to test this class.

tests/System/env/config_dist/phpbench.json.dist
```diff
@@ -1,4 +1,4 @@
 {
-    "bootstrap": "../../../../../vendor/autoload.php",
+    "bootstrap": "../../../../vendor/autoload.php",
```
dantleech: Why did this change?

tests/Unit/Benchmark/Metadata/FactoryTest.php~
```diff
@@ -0,0 +1,248 @@
+<?php
+
```
dantleech: remove this.

lib/Reflection/ComposerFileReflector.php
dantleech: At least add a comment

PR #430, opened 2016-07-08
Adjust lstrojny/functional-php dependency version

elazar:
Allow this dependency to be updated while avoiding a compatibility issue with
HHVM present in lstrojny/functional-php:1.2.1. This issue is fixed in 1.2.3.


dantleech:
Test failure is because of the PHAR builder. Will merge and revert if it causes issues, thanks :)

PR #449, opened 2016-10-02
Allow custom subject pattern

dantleech:
Allow use of custom pattern for matching benchmark subjects.


Hywan:
Can we help?

dantleech:
@Hywan i think this one is pretty simple, feel free to review if you like however :)

PR #471, opened 2017-04-18
Remove dead code

villfa:


dantleech:
Thanks, this is actually a dead method - its not used by PHPBench itself, would be great if you can add a test for it or perhaps better - we could remove it as the stat can be accessed via. the Results anyway.

villfa:
@dantleech You're right, this method was not used anywhere so I removed it.

dantleech:
Awesome., thanks :)

PR #474, opened 2017-07-07
Assertions

dantleech:
Introduce assertions.

image

Assertions are specified as follows (the comparator defaults to less than):

```php
/**
 * @Assert(stat="mean", value=100)
 */
public function benchSomethingLess()

/**
 * @Assert(stat="mean", value=100, comparator=">")
 */
public function benchSomethingGreater()
```
In addition assertions can be globally overridden on the CLI:

```
$ phpbench run --assert="stat: mean, value: 100"
```

Warnings are issued if the `tolerance` parameter is given:

```php
/**
 * @Assert(stat="mean", value=100, tolerance=10)
 */
public function benchSomethingGreater()
```

In this case if the value is 105, a warning will be issued but PHPBench will exit with 0.

TODO:

- [x] Serialize / deserialize assertion failures.
- [x] Enable failures in all progress loggers.
- [x] Exit with a non-zero code if there were failures.
- [x] Show list of failures in summary (similar to when exceptions are encountered
- [x] Respect precision (e.g.` was greater than 12.123123123123`)
- [x] Allow time unit specification in Comparator asserter
- [x] Allow use of throughput for assertions
- [x] Allow tolerance threshold? I.e. only WARN if its within the threshold.
- [ ] Allow selection of asserter `@Assert("blah", asserter="foobar")`

BUGS

- [x] Assertions are appended when used with retry threshold


composer.json
```diff
@@ -26,7 +26,8 @@
         "ext-json": "*",
         "ext-pcre": "*",
         "ext-reflection": "*",
-        "ext-spl": "*"
+        "ext-spl": "*",
+        "hoa/ruler": "2.17.05.16"
```
Hywan: I would recomend `~2.0` or at least `~2.17` but it does not matter.

lib/Benchmark/Asserter/RulerAsserter.php
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    public function assert(string $expression, Distribution $distribution)
+    {
+        $context = new Context();
+        foreach ($distribution->getStats() as $key => $value) {
+            $context[$key] = $value;
+        }
+
+        if (false === $this->ruler->assert($expression, $context)) {
```
Hywan: :heart:

lib/Benchmark/AssertionFailure.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace PhpBench\Benchmark;
+
+class AssertionFailure
+{
+    private $message;
+
+    public function __construct($message)
```
Hywan:
```diff
- public function __construct($message)
+ public function __construct(string $message)
```

lib/Benchmark/Metadata/Annotations/Assert.php
```diff
@@ -0,0 +1,36 @@
...
+ * @Taget({"METHOD", "CLASS"})
+ * @Attributes({
+ *    @Attribute("value", required = true, type="string"),
+ * })
+ */
+class Assert extends AbstractArrayAnnotation
+{
+    private $expression;
+
+    public function __construct($params)
```
Hywan:
```diff
- __construct($params)
+ __construct(array $params)
```

lib/Benchmark/Metadata/AssertionMetadata.php
```diff
@@ -0,0 +1,21 @@
...
+namespace PhpBench\Benchmark\Metadata;
+
+class AssertionMetadata
+{
+    private $expression;
+
+    /**
+     * @param string $expression
+     */
+    public function __construct($expression)
```
Hywan:
```diff
- __construct($expression)
+ __construct(string $expression)
```

And thus, you can remove the doc comment.

lib/Extension/CoreExtension.php
```diff
@@ -74,6 +74,8 @@
 use PhpBench\Util\TimeUnit;
 use Symfony\Component\Finder\Finder;
 use Symfony\Component\Process\ExecutableFinder;
+use Hoa\Ruler\Ruler;
```
Hywan: Is it used in this file?

lib/Model/Variant.php
```diff
@@ -355,6 +362,16 @@ public function setException(\Exception $exception)
         $this->errorStack = new ErrorStack($this, $errors);
     }
 
+    public function addFailure(AssertionFailure $failure)
+    {
+        $this->failures[] = $failure;
+    }
+
+    public function hasFailed()
```
Hywan:
```diff
- hasFailed()
+ hasFailed(): bool
```

Hywan:
And thank you for using Hoa!

lib/Benchmark/AssertionFailure.php
dantleech: PHPBench is still 5.6 compatible currently

dantleech:
Thanks for the feedback, early stage PR :)

composer.json
dantleech: Hmm, thats just what composer gave me, will use `~2.0`

lib/Benchmark/AssertionFailure.php
dantleech: Maybe it's time to move to 7 though

Hywan:
@dantleech Early feedbacks too :-p.

dantleech:
I am trying the Symfony Expression Lanaguage and I feel like a bit of a troll :/ In this context both the Symfony EL and Hoa Ruler are functionally equivalent (as far as I can tell). 

As this project is already heavily based on the Symfony stack, it seems more natural to use the Symfony component rather than to bring in the Hoa component (and bring in several Hoa dependencies). In a different world it would be based on the Hoa stack, but things are as they are.

Sorry for Tweeting about the Hoa usage prematurely.

Hywan:
I understand your arguments. That's OK for me ;-).

lib/Assertion/AsserterRegistry.php
```diff
@@ -0,0 +1,27 @@
...
+use PhpBench\Registry\Registry;
+
+/**
+ * @method \PhpBench\Assertion\Asserter getService()
+ */
+class AsserterRegistry extends Registry
+{
+    public function __construct(Container $container, string $defaultService = null)
+    {
+        parent::__construct('assertion', $container, $defaultService);
```
dantleech: constant

lib/Benchmark/Metadata/Annotations/Assert.php
```diff
@@ -0,0 +1,36 @@
...
+     */
+    private $options;
+
+    public function __construct($params)
+    {
+        parent::__construct($params);
+        $this->options = $params;
+    }
+
+    public function getOptions(): array
```
dantleech: Rename => config

lib/Benchmark/Metadata/AssertionMetadata.php
```diff
@@ -0,0 +1,31 @@
...
+class AssertionMetadata
+{
+    /**
+     * @var array
+     */
+    private $options;
+
+    public function __construct(array $options)
+    {
+        $this->options = $options;
```
dantleech: rena,e

lib/Benchmark/Runner.php
```diff
@@ -139,6 +163,15 @@ private function runBenchmark(
             $subjectMetadata->setSleep($context->getSleep($subjectMetadata->getSleep()));
             $subjectMetadata->setRetryThreshold($context->getRetryThreshold($this->retryThreshold));
 
+            if ($context->getAssertions()) {
+                $subjectMetadata->setAssertions($this->assertionProcessor->assertionsFromRawCliConfig($context->getAssertions()));
+            }
+
+            if ($context->getAssertions()) {
+                foreach ($context->getAssertions() as $assertion) {
+                }
```
dantleech: remove

lib/Model/Suite.php
```diff
@@ -151,6 +151,22 @@ public function getErrorStacks()
         return $errorStacks;
     }
 
+    public function getFailures()
```
dantleech: `@return AssertionFailures[]`

dantleech:
This is now pretty much done - in the end I decided not to use expressions at all, but it is written in such a way that they could be introduced in the future as an extension or part of the core.]

The default asserter in this PR has the following form:

```
@Assert(stat="mean", value=100, comparator=">")
```

But writing an expression asserter would allow:

```
@Assert("mean > 100", asserter="ruler")
```

(and of course the default asserter _could_ be defined in config).

lib/Assertion/ComparatorAsserter.php
```diff
@@ -0,0 +1,193 @@
...
+            TimeUnit::MODE_TIME,
+            $timeUnit,
+            $comparator,
+            $tolerance
+        );
+    }
+
+    private function check(string $failureMessage, $value, $expectedValue, string $statName, string $mode, string $timeUnit, string $comparator, $tolerance)
+    {
+        if (false === $this->compare($expectedValue, $value, $comparator)) {
```
dantleech: Use early retrn

lib/Assertion/ComparatorAsserter.php
```diff
@@ -0,0 +1,193 @@
...
+
+    private function formatValue($value, $timeUnit, $mode)
+    {
+        switch ($mode) {
+            case TimeUnit::MODE_THROUGHPUT:
+                return $this->formatThroughput($value, $timeUnit);
+            case TimeUnit::MODE_TIME:
+                return $this->timeUnit->format($value, $timeUnit, TimeUnit::MODE_TIME);
+        }
+    }
```
dantleech: Runtime exception if it gets this far (which it should not).

PR #475, opened 2017-07-15
Optimized microtime template

marc-mabe:
... to not use variable function call and use a compiled number of revolutions.


dantleech:
Well spotted, thanks. The Travis build doesn't work for thrid-party PRs at the moment, hopefully it will pass in master :)

PR #476, opened 2017-07-23
Allow Symfony 4 components as dependency

lcobucci:
So that we can use this project on https://github.com/doctrine/doctrine2


dantleech:
Thanks :)

lcobucci:
@dantleech thank you too 😉

PR #482, opened 2017-10-09
Fix: Actually use JSON in configuration example

localheinz:
This PR

* [x] actually uses JSON in a configuration example


dantleech:
:+1: approved

localheinz:
Thank you, @dantleech!

PR #483, opened 2017-10-09
Enhancement: Keep packages sorted in composer.json

localheinz:
This PR

* [x] keeps packages sorted in `composer.json`

💁‍♂️ For reference, see https://getcomposer.org/doc/06-config.md#sort-packages.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,31 +8,34 @@
             "email": "daniel@dantleech.com"
         }
     ],
+    "config": {
+        "sort-packages": true
+    },
     "require": {
         "php": "^7.0",
-        "phpbench/dom": "~0.2.0",
-        "phpbench/container": "~1.0",
-        "symfony/console": "^2.6|^3.0|^4.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -8,31 +8,34 @@
             "email": "daniel@dantleech.com"
         }
     ],
+    "config": {
+        "sort-packages": true
```
localheinz: After adding this configuration, I ran

```
$ composer require php:^7.0
```

and

```
$ composer require --dev phpunit/phpunit:^6.0
```

to trigger sorting.

localheinz:
@dantleech 

Build is failing, to be honest, I have no idea how that could be related to the proposed changes!

dantleech:
Build always fails for 3rd party PRs, it's realted to the PHAR build. Will merge and hope for the best :)

dantleech:
Can always revert it

localheinz:
Thank you, @dantleech!

PR #484, opened 2017-10-09
Fix: Install dependencies in install section

localheinz:
This PR

* [x] uses the `install` section for installing dependencies on Travis CI

💁‍♂️ For reference, see https://docs.travis-ci.com/user/customizing-the-build#The-Build-Lifecycle.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,11 @@ cache:
 before_install:
     - composer validate
 
-before_script:
+install:
     - if [ -n "${SYMFONY_VERSION}" ]; then composer require symfony/symfony:${SYMFONY_VERSION} --no-update; fi;
     - composer update ${COMPOSER_FLAGS} --prefer-dist
+
+before_script:
     - export TZ=Europe/Paris
 
 script:


.travis.yml
```diff
@@ -6,16 +6,19 @@ php:
 
 sudo: false
 
+env:
+  global:
+    - TZ=Europe/Paris
```
localheinz: See https://docs.travis-ci.com/user/environment-variables/.

.travis.yml
```diff
@@ -6,16 +6,19 @@ php:
...
+
 cache:
   directories:
     - "$HOME/.composer/cache"
 
-before_script:
+install:
```
localheinz: See https://docs.travis-ci.com/user/customizing-the-build#The-Build-Lifecycle.

.travis.yml
localheinz: Also see this in effect here: https://travis-ci.org/phpbench/phpbench/jobs/285491568#L429-L430.

.travis.yml
localheinz: Since we install dependencies here, it makes more sense to use the `install` section rather than `before_script`. 

There's another reason I will get to in another PR, hehe! 🤓

.travis.yml
dantleech: Can't hurt, but what is the advantage?

.travis.yml
dantleech: :+1:

.travis.yml
localheinz: In my opinion, the advantage is that it's more obviously stated that we define an environment variable used in the builds.

What do you think?

dantleech: I guess I'd prefer to leave it out unless the TZ is having a beneficial effect on the build, looking at the Symfony `.travis.yml` they don't seem to include this

localheinz: No problem, let me sort this out later.

dantleech: Cool, thanks for the improvements so far :)

localheinz:
Closing in favour of #570.

PR #486, opened 2017-10-09
Fix: Case mismatch in method call or class name

localheinz:
This PR

* [x] fixes case mismatches in method calls or class names

### DIFF
--- a/lib/Report/Generator/Table/Sort.php
+++ b/lib/Report/Generator/Table/Sort.php
@@ -35,8 +35,8 @@ class Sort
         $array2 = array_slice($array, $halfway);
 
         // Recurse to sort the two halves
-        self::mergesort($array1, $callback);
-        self::mergesort($array2, $callback);
+        self::mergeSort($array1, $callback);
+        self::mergeSort($array2, $callback);
 
         // If all of $array1 is <= all of $array2, just append them.
         if (call_user_func($callback, end($array1), $array2[0]) < 1) {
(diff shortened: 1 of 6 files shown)


lib/Report/Generator/Table/Sort.php
```diff
@@ -35,8 +35,8 @@ public static function mergeSort(&$array, \Closure $callback)
         $array2 = array_slice($array, $halfway);
 
         // Recurse to sort the two halves
-        self::mergesort($array1, $callback);
-        self::mergesort($array2, $callback);
+        self::mergeSort($array1, $callback);
+        self::mergeSort($array2, $callback);
```
localheinz: See [`PhpBench\Report\Generator\Table\Sort::mergeSort()`](https://github.com/phpbench/phpbench/blob/da9dbd7a690789463cfe9c37c19a732dba872045/lib/Report/Generator/Table/Sort.php#L25-L66).

lib/Report/Renderer/XsltRenderer.php
```diff
@@ -97,7 +97,7 @@ public function render(Document $reportDom, Config $config)
 
         $stylesheetDom = new \DOMDocument('1.0');
         $stylesheetDom->load($template);
-        $xsltProcessor = new \XsltProcessor();
+        $xsltProcessor = new \XSLTProcessor();
```
localheinz: See http://php.net/manual/en/class.xsltprocessor.php.

tests/System/SystemTestCase.php
```diff
@@ -105,7 +105,7 @@ protected function assertXPathCount($count, $xmlString, $query)
     protected function assertXPathExpression($expected, $xmlString, $expression)
     {
         $dom = new \DOMDocument();
-        $result = @$dom->loadXml($xmlString);
+        $result = @$dom->loadXML($xmlString);
```
localheinz: See http://php.net/manual/en/domdocument.loadxml.php.

localheinz:
Thank you, @dantleech!

PR #487, opened 2017-10-09
Fix: Use more appropriate assertions

localheinz:
This PR

* [x] uses more appropriate assertions in tests


dantleech:
:+1:

localheinz:
Thank you, @dantleech!

PR #490, opened 2017-10-15
Disable php ini

dantleech:
Added option to disable PHP ini:

- [x] Added `--php-disable-ini`
- [x] Added list of loaded extensions and path to INI to the PHP env provider.
- [x] Fixed the `php-config` option to allow multiple values, (e.g. `extension: [ json.so, phar.so ]`).

**NOTE**: Unforfortunately PHPBench currently requires the JSON extension to be loaded in the remote process - it has to be explicitly enabled, we could automatically enable it, but this might causes WTFs. Currently we just throw a helpful exception.

I think in the future we should change this and the remote processes should just return formatted plain text.

tmp

So a configuration might look like this:

```javascript
{
    "bootstrap": "vendor/autoload.php",
    "path": "benchmarks",
    "php_disable_ini": true,
    "php_config": {
        "extension": [ "json.so" ]
    }
}
```

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -233,14 +233,14 @@ The above will allow you to have benchmark class such as:
 Disable the PHP INI file
 ------------------------
 
-PHP extensions, especially XDebug can drastically affect the performance of
-your benchmarks. You can disable XDebug and other dynamically loaded
+PHP extensions, especially XDebug, can drastically affect the performance of
+your benchmark subjects. You can disable XDebug and other dynamically loaded
 extensions by setting ``php_disable_ini`` to ``true``.
 
 .. note:
 
-    PHPBench currently makes use of the ``json`` extension remote
-    processes, so you are required to explicitly enable this as follows.
+    PHPBench currently makes use of the ``json`` extension in remote
+    processes, so you are required to explicitly enable it as follows.
 
 .. code-block:: javascript
 
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -227,3 +227,26 @@ The above will allow you to have benchmark class such as:
 
     You can also explicitly declare that methods are benchmark subjects by
     using the ``@Subject`` annotation.
+
+.. _configuration_disable_php_ini:
+
+Disable the PHP INI file
+------------------------
+
+PHP extensions, especially XDebug can drastically affect the performance of
+your benchmarks. You can disable XDebug and other dynamically loaded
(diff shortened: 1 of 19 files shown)


lib/Extension/CoreExtension.php
```diff
@@ -484,7 +486,7 @@ public function registerEnvironment(Container $container)
         $container->register('environment.provider.php', function (Container $container) {
             return new Provider\Php(
                 $container->get('benchmark.remote.launcher'),
-                $container->getParameter('php_binary') !== PHP_BINARY && $container->getParameter('php_binary') !== null
+                true // always use remote env information
```
dantleech: dd

lib/PhpBench.php
```diff
@@ -113,6 +113,10 @@ private static function loadConfig()
                 $value = $jsonParser->decode($value);
                 $configOverride['php_config'] = $value;
             }
+
+            if ($arg == '--php-disable-ini') {
+                $configOverride['php_disable_ini'] = true;
```
dantleech: Add integration test

docs/configuration.rst
```diff
@@ -227,3 +227,26 @@ The above will allow you to have benchmark class such as:
 
     You can also explicitly declare that methods are benchmark subjects by
     using the ``@Subject`` annotation.
+
+.. _configuration_disable_php_ini:
+
+Disable the PHP INI file
+------------------------
+
+PHP extensions, especially XDebug can drastically affect the performance of
```
dantleech: missing comma.

docs/configuration.rst
```diff
@@ -227,3 +227,26 @@ The above will allow you to have benchmark class such as:
...
+your benchmarks. You can disable XDebug and other dynamically loaded
+extensions by setting ``php_disable_ini`` to ``true``.
+
+.. note:
+
+    PHPBench currently makes use of the ``json`` extension remote
```
dantleech: extension _in_ remote ...

lib/Benchmark/Executor/BaseExecutor.php
```diff
@@ -69,6 +69,11 @@ public function execute(SubjectMetadata $subjectMetadata, Iteration $iteration,
         return $this->launch($payload, $iteration, $config);
     }
 
+    public function healthCheck()
+    {
+        $result = $this->launcher->payload(__DIR__ . '/template/health_check.template', [])->launch();
```
dantleech: Remove assignmentb

PR #492, opened 2017-10-15
Introduced Opcode env provider

dantleech:
Simply show if the opcache is enabled in the benchmarking environment:

asd


lib/Environment/Information.php
```diff
@@ -94,4 +94,21 @@ public function getIterator()
...
+
+            if (is_array($value)) {
+                $transformed = array_merge($transformed, $this->flattenInformation($value, $key));
+                continue;
+            }
+
+            $transformed[$key] = $value;
+        }
+
+        return $transformed;
```
dantleech: Add test for this ...

PR #498, opened 2017-11-12
Windows filename handling and commandline escaping

RaoulSchaffranek:
Hi,

i discovered two bugs when trying to run phpbench from Windows 10:

1. When resolving the benchmark-directory phpbench prepends the current working directory to relative paths. Whether or not a path is absolute is determined by checking if the first character of the path is a forward slash. On Windows hosts, however, absolute paths start with a drive-name. Therefore the current working directory was prepended for absolute paths as well.

2. On Windows hosts the php-binary usually lives under `C:\Program Files\`. When starting a new process phpbench must account for the space in the filename and wrap the path of the binary in double-quotes.

With these two fixes i was able to run phpbench successfully under my Windows 10 system.

### DIFF
--- a/lib/Benchmark/Remote/Payload.php
+++ b/lib/Benchmark/Remote/Payload.php
@@ -70,6 +70,10 @@ class Payload
      */
     public function __construct($template, array $tokens = [], $phpBinary = PHP_BINARY, Process $process = null)
     {
+        if (PHP_OS === 'WINNT') {
+            $phpBinary = sprintf('"%s"', $phpBinary);
+        }
+
         $this->template = $template;
         $this->process = $process ?: new Process($phpBinary);
         $this->tokens = $tokens;
--- a/lib/Benchmark/Remote/Payload.php
+++ b/lib/Benchmark/Remote/Payload.php
@@ -70,17 +70,13 @@ class Payload
      */
     public function __construct($template, array $tokens = [], $phpBinary = PHP_BINARY, Process $process = null)
     {
-        if (PHP_OS === 'WINNT') {
-            $phpBinary = sprintf('"%s"', $phpBinary);
-        }
-
+        $this->phpBinary = escapeshellarg($phpBinary);
         $this->template = $template;
-        $this->process = $process ?: new Process($phpBinary);
+        $this->process = $process ?: new Process($this->phpBinary);
         $this->tokens = $tokens;
 
         // disable timeout.
         $this->process->setTimeout(null);
-        $this->phpBinary = $phpBinary;
         $this->iniStringBuilder = new IniStringBuilder();
     }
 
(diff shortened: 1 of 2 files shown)


lib/Benchmark/Remote/Payload.php
```diff
@@ -70,6 +70,10 @@ class Payload
      */
     public function __construct($template, array $tokens = [], $phpBinary = PHP_BINARY, Process $process = null)
     {
+        if (PHP_OS === 'WINNT') {
+            $phpBinary = sprintf('"%s"', $phpBinary);
```
dantleech: I think it would be better to do this?

```php
$this->phpBinary = escapeshellarg($phpBinary)
```

?

lib/PhpBench.php
```diff
@@ -73,6 +73,10 @@ public static function normalizePath($path)
             return $path;
         }
 
+        if (PHP_OS === 'WINNT') {
+            return stream_resolve_include_path($path);
```
dantleech: I think instead of checking to see if the first char is `/` we should use the [webmozart/path-util](https://packagist.org/packages/webmozart/path-util) package:

```
if (Path::isAbsolute($path)) {
    return $path;
}
```

dantleech:
Thanks for the PR! Left a couple of comments, should really think about adding Appveyor for testing on Windows.

lib/Benchmark/Remote/Payload.php
RaoulSchaffranek: I first tried escapeshellcmd and escapeshellarg, unfortunately both failed so i went with this manual solution.

lib/PhpBench.php
RaoulSchaffranek: I don't know the purpose of `normalizePath` well enough, but i guess it can be safely replaced by `stream_resolve_include_path`. At least, that's how phpunit resolves its test-direcotries.

dantleech: Well, it needs to make relative paths relative to where the command is executed. Without checking I am not sure to how many contexts this applies. Is there any use case for including things relative to all include paths?

If not I would prefer to just use the PathUtil fix above.

RaoulSchaffranek: I cannot think of any non-pathological use case. However, i'd say that `stream_resolve_include_path` produces the least surprising results because path-resolution then works in exactly the same way as for `require`, `fopen`, `file`, etc.

lib/Benchmark/Remote/Payload.php
RaoulSchaffranek: Sorry, i just double-checked my former statement and `escapeshellarg` works like a charm. I cleaned that up.

lib/PhpBench.php
dantleech: The reason for using `getcwd` was to solve issues with resolving paths with the PHAR distribution. I will have to test the PHAR build with `stream_resolver_include_path`, and check exactly how PHPUnit does it.  However, I am still not sure if we want to be including things from all the include paths.

So, I am still leaning towards using the Path library as suggested above and keeping the existing `getcwd` logic.

dantleech:
Have picked your commits into #507 and made some changes, thanks for the PR!

PR #502, opened 2017-12-03
Optimization flags added for the application

Spomky:
This modification adds the following flags for the CLI application:

- `o`: Optimize autoloader
- `a`: Autoload classes from the classmap only

See https://getcomposer.org/doc/articles/autoloader-optimization.md for additional information.


dantleech:
Thanks! The build currently always fails for 3rd party PRs.

PR #507, opened 2018-01-03
Windows paths

dantleech:
Replaces #498


dantleech:
@fhein @RaoulSchaffranek can you confirm that this works on windows now?

fhein:
I have not tried to use it with UNC names of network resources like \\server\share\directory\file.php. But for local absolute paths like C:\dir\file.php it works.

fhein:
It works if you give run an argument (the directory or file). If you omit the argument, the same error as of #498.

PR #510, opened 2018-01-05
Fix: Remove unnecessary parentheses

localheinz:
This PR

* [x] remove unnecessary parentheses when using `require`


localheinz:
Thank you, @dantleech!

dantleech:
Why not ... :)

PR #511, opened 2018-01-05
Fix: Case mismatch

localheinz:
This PR

* [x] fixes case mismatches when invoking methods


localheinz:
Thank you, @dantleech!

PR #515, opened 2018-02-11
Exclude is not a valid option anymore

sagikazarmark:


dantleech:
Actually never was - but something like it would be great ....

PR #522, opened 2018-03-07
Allowing NULL values using ParamProviders

popnikos:
While first attempt to bench something using the @ParamProviders annotation...I got InvalidArgumentException messages such as : 
    Parameter values must be scalar. Got "NULL" 
    Parameters must be either scalars or arrays, got: NULL

But I want my bench method to have null values into array of parameters

So I change few lines into the code and all went as excpected.

I have not check further impact in my modifications but I really need this feature.

### DIFF
--- a/lib/Serializer/XmlEncoder.php
+++ b/lib/Serializer/XmlEncoder.php
@@ -186,7 +186,6 @@ class XmlEncoder
     {
         $parameterEl = $parentEl->appendElement('parameter');
         $parameterEl->setAttribute('name', $name);
-
         if (is_array($value)) {
             $parameterEl->setAttribute('type', 'collection');
             foreach ($value as $key => $element) {
@@ -195,19 +194,19 @@ class XmlEncoder
 
             return $parameterEl;
         }
-
+        if (is_null($value)) {
+            $parameterEl->setAttribute('xsi:nil', 'true');
+            
+            return $parameterEl;
+        }
+        
         if (is_scalar($value)) {
             $parameterEl->setAttribute('value', $value);
 
             return $parameterEl;
         }
 
-        if (is_null($value)) {
-            $parameterEl->setAttribute('nillable', 'true');
-            $parameterEl->setAttribute('value', '');
-
-            return $parameterEl;
-        }
+        
 
         throw new \InvalidArgumentException(sprintf(
             'Parameters must be either scalars or arrays, got: %s',
--- a/lib/Serializer/XmlEncoder.php
+++ b/lib/Serializer/XmlEncoder.php
@@ -201,7 +201,7 @@ class XmlEncoder
 
             return $parameterEl;
         }
-        
+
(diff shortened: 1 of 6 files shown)


lib/Serializer/XmlEncoder.php
```diff
@@ -202,6 +202,13 @@ private function createParameter($parentEl, $name, $value)
             return $parameterEl;
         }
 
+        if (is_null($value)) {
+            $parameterEl->setAttribute('nillable', 'true');
+            $parameterEl->setAttribute('value', '');
+
+            return $parameterEl;
+        }
```
dantleech: May need to modify the `XmlDecoder` also

dantleech:
> Can you check the failed message

This is a known issue with 3rd party contributions and Travis at the moment, you can ignore it (just ensure the tests run locally).

I think this feature is fine, but we should add:

- [ ] A test case in `tests/Unit/Benchmark/Remote/ReflectorTest.php`
- [ ] A test case in `tests/Unit/Serlializer/XmlTestCase.php` (should be enough to add a case to `provideEncode`).

And then will probably need to modify the XmlDecoder based on that.

lib/Serializer/XmlEncoder.php
```diff
@@ -202,6 +202,13 @@ private function createParameter($parentEl, $name, $value)
             return $parameterEl;
         }
 
+        if (is_null($value)) {
+            $parameterEl->setAttribute('nillable', 'true');
```
dantleech: Would prefer `nullable`, what do you think?

lib/Serializer/XmlEncoder.php
popnikos: I've had a quick look and found no impact. but i will review it ASAP

lib/Serializer/XmlEncoder.php
popnikos: regarding xml uses i will change it to <code>xsi:nil</code>
https://stackoverflow.com/questions/774192/what-is-the-correct-way-to-represent-null-xml-elements

popnikos:
Hi @dantleech I've done some stuff to complete the PR.
For now I can't run PHP Unit on my computer (due to really specific configuration), and as Travis failed, I'm not sure tests will pass
Does anyone can check if tests are OK or not ?
Noticed : It still remains Reflector test (I don't have the time right now I will try this week-end)

popnikos:
I've finally added test case in Reflector.
Let know if you can run test on your side while I can't run for now.

dantleech:
Great! I will try and check it out this evening :)

popnikos:
I have succeeded  with tests ... I've set up a dockerfile to get php7 instead of my local php5.6 version
I've updated the source.
Let me know if you need other update or if you validate the PR 

Thank you @dantleech

dantleech:
Tested locally, and it's fine. Thanks for your effort :)

PR #536, opened 2018-05-13
Callable param provider

dantleech:
- [x] Allow any callable as a param provider
- [x] Made CHANGELOG an `.md` file.


lib/Benchmark/Remote/template/parameter_set_extractor.template
```diff
@@ -22,7 +22,19 @@ $parameterSets = array();
 ob_start();
 
 foreach ($paramProviders as $paramProvider) {
-    $parameterSets[] = $benchmark->$paramProvider();
+    if (is_callable($paramProvider)) {
+        $parameterSets[] = call_user_func($paramProvider);
+        continue;
+    }
+
+    if (method_exists($class, $paramProvider)) {
```
dantleech: Should be the other way round - local method to take precedence over global function f.e.

rybakit: This check will give false positives for non-public methods: https://3v4l.org/j5kOr.

lib/Benchmark/Remote/template/parameter_set_extractor.template
```diff
@@ -22,7 +22,19 @@ $parameterSets = array();
 ob_start();
 
 foreach ($paramProviders as $paramProvider) {
-    $parameterSets[] = $benchmark->$paramProvider();
+    if (is_callable($paramProvider)) {
+        $parameterSets[] = call_user_func($paramProvider);
```
rybakit: Can be simplified/optimized to `$parameterSets[] = $paramProvider();`

lib/Benchmark/Remote/template/parameter_set_extractor.template
dantleech: I guess we could normalize it to `$className . '::' . $paramProvider` https://3v4l.org/YnBCN

rybakit: Yep, either `is_callable("$className::$paramProvider")` or `is_callable([$benchmark, $paramProvider])` will work.

CHANGELOG.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Features
+
+- Assertions [docs](http://phpbench.readthedocs.io/en/latest/writing-benchmarks.html#assertions)
+- Added `--disable-php-ini` option
+- Report if opcode extension is enabled in env.
+- Show enabled PHP extensions in env.
+- PHP 7 only
+
+## Bugfixes
```
dantleech: should be `###`

benchmarks/Micro/HashingBench.php
```diff
@@ -12,24 +12,27 @@
...
-    }
-
-    public function benchSha256()
+    /**
+     * @ParamProviders({"\PhpBench\Benchmarks\Micro\hash_algos"})
+     */
+    public function benchAlgos($params)
     {
-        return hash('sha256', 'hello world');
+        return md5($params[0]);
```
dantleech: md5?

dantleech:
:boom:

PR #562, opened 2018-11-11
Refactor CartesianParameterIterator: Depend on CartesianIterator

PatchRanger:
Didn't test it thoroughly, did just a few smoke-tests, looks good to me.
The idea is to get rid of self-crafted implementation of Cartesian product, relying on "mine-crafted" (read as "crafted by me") version of CartesianIterator, which in turn relies on standard, built-in [`MultipleIterator`](http://php.net/manual/en/class.multipleiterator.php), look how laconic it is: https://github.com/PatchRanger/cartesian-iterator/blob/master/src/CartesianIterator.php .
Could bring performance boost - but not for sure.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,6 +21,7 @@
         "beberlei/assert": "^2.4|^3.0",
         "doctrine/annotations": "^1.2.7",
         "lstrojny/functional-php": "1.0|^1.2.3",
+        "patchranger/cartesian-iterator": "~0.0.5",
         "phpbench/container": "~1.2",
         "phpbench/dom": "~0.2.0",
         "seld/jsonlint": "^1.0",
--- a/lib/Benchmark/CartesianParameterIterator.php
+++ b/lib/Benchmark/CartesianParameterIterator.php
@@ -12,51 +12,34 @@
 
 namespace PhpBench\Benchmark;
 
+use PatchRanger\CartesianIterator;
 use PhpBench\Model\ParameterSet;
 
-class CartesianParameterIterator implements \Iterator
+class CartesianParameterIterator extends CartesianIterator
 {
-    private $sets;
     private $index = 0;
-    private $max;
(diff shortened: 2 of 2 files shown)


PatchRanger:
I haven't fully debugged yet: some tests still fail.
I decided to run a benchmark to see whether my proposal is better in terms of efficiency. Here are the results:

Original

benchmark | subject | groups | params | revs | its | mem_peak | best | mean | mode | worst | stdev | rstdev | diff
-|-|-|-|-|-|-|-|-|-|-|-|-|-
RunBench | benchRun | | [] | 1 | 10 | 5,483,640b | 0.307s | 0.312s | 0.314s | 0.316s | 0.003s | 1.03% | 1.00x
RunBench | benchRunAndReport | | [] | 1 | 10 | 5,838,805b | 0.311s | 0.317s | 0.318s | 0.321s | 0.004s | 1.14% | 1.02x

PR

benchmark | subject           | groups | params | revs | its | mem_peak   | best   | mean   | mode   | worst  | stdev  | rstdev | diff
-|-|-|-|-|-|-|-|-|-|-|-|-|-
RunBench  | benchRun          |        | [[]]   | 1    | 10  | 5,490,450b | 0.324s | 0.427s | 0.369s | 0.789s | 0.136s | 31.92% | 1.00x
RunBench  | benchRunAndReport |        | [[]]   | 1    | 10  | 5,845,747b | 0.339s | 0.431s | 0.373s | 0.651s | 0.102s | 23.67% | 1.01x

It definitely looks like my implementation is less effective.
I guess I should close the PR.

dantleech:
Thanks for the effort, it would indeed be nice to replace some of PHPBench's internals with better implementations, but would rather do this with established libraries, for example https://github.com/markrogoyski/math-php - but if it ain't broke, don't fix it :)

PR #565, opened 2018-12-24
Fix: Keep rules sorted in .php_cs.dist

localheinz:
This PR

* [x] keeps rules sorted in `.php_cs.dist`


localheinz:
Thank you, @dantleech!

PR #566, opened 2018-12-24
Enhancement: Update phpstan/phpstan

localheinz:
This PR

* [x] updates `phpstan/phpstan`
* [x] configures paths to analyse in `phpstan.neon`
* [x] configures `phpstan` level via `includes` in `phpstan.neon`
* [x] uses 🇬🇧 spelling when asking `phpstan` to analyse code
* [x] removes an ignored error that does not match any reported errors anymore
* [x] fixes issues detected by `phpstan`
* [x] removes ignored error not matched anymore
* [x] ignores errors for now
* [x] sets terminal dimensions via environment variables as `setTerminalDimensions()` has been removed in `symfony/console:^4.0.0`

💁‍♂️ For reference, see https://github.com/phpstan/phpstan/compare/0.8.5...0.10.6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ jobs:
         - composer install
 
       script:
-        - ./vendor/bin/phpstan analyze -c phpstan.neon
+        - ./vendor/bin/phpstan analyse -c phpstan.neon
 
     - &TEST
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ jobs:
         - composer install
 
       script:
-        - ./vendor/bin/phpstan analyze --level=4 -c phpstan.neon
+        - ./vendor/bin/phpstan analyze -c phpstan.neon
 
     - &TEST
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ jobs:
         - composer install
 
       script:
-        - ./vendor/bin/phpstan analyze --level=4 -c phpstan.neon lib/
+        - ./vendor/bin/phpstan analyze --level=4 -c phpstan.neon
 
     - &TEST
 
--- a/lib/Benchmark/Remote/Launcher.php
+++ b/lib/Benchmark/Remote/Launcher.php
@@ -62,7 +62,13 @@ class Launcher
     private $phpDisableIni;
 
     /**
-     * @param mixed string
+     * @param PayloadFactory|null $factory
+     * @param ExecutableFinder|null $finder
+     * @param string|null $bootstrap
+     * @param string|null $phpBinary
+     * @param array $phpConfig
+     * @param null|string $phpWrapper
+     * @param bool $phpDisableIni
      */
     public function __construct(
         PayloadFactory $factory = null,
@@ -132,6 +138,7 @@ class Launcher
         }
 
         // otherwise try and find it in PATH etc.
+        /** @var string|null $phpBinary */
         $phpBinary = $this->finder->find($this->phpBinary);
 
         if (null === $phpBinary) {
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -3,7 +3,6 @@ includes:
 
 parameters:
     ignoreErrors:
-        - '#Undefined variable: \$argv#'
         - '#Call to an undefined method DOMNode::#'
     paths:
         - lib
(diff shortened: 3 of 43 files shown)


phpstan.neon
```diff
@@ -1,3 +1,5 @@
 parameters:
     ignoreErrors:
-        - '#Undefined variable: \$argv#'
+        - '#Variable \$reject might not be defined.#'
```
localheinz: As far as `phpstan/phpstan` is concerned, it's right: the variable might not be defined.

Other than that, I think it is safe to assume that it is defined, see https://github.com/phpbench/phpbench/blob/f72b3446c351d8f7ac82a41b30892739890b4d72/lib/Benchmark/Runner.php#L277-L280

.travis.yml
```diff
@@ -43,7 +43,7 @@ deploy:
         branch: v1.8.47
     provider: pages
     skip-cleanup: true
-    github-token:
```
localheinz: Unrelated change, but seems that whitespace at  line-endings wasn't previously trimmed.

dantleech:
Looks like we accidentally hit a BC break in the Symfony console component, `Application::setTerminalDimenstions` doesn't exist :confused:

tests/Unit/Console/Command/LogCommandTest.php
```diff
@@ -146,7 +152,8 @@ public function testPaginate()
         $input = new ArrayInput([], $this->command->getDefinition());
         $output = $this->output;
 
-        $this->application->setTerminalDimensions(100, 14);
+        $this->setTerminalDimensions(100, 14);
```
dantleech: might be better to keep this at the application level (see failing tests)?

lib/Benchmark/Remote/Launcher.php
```diff
@@ -131,6 +137,7 @@ private function resolvePhpBinary()
         }
 
         // otherwise try and find it in PATH etc.
+        /** @var string|null $phpBinary */
```
localheinz: Perhaps this should be amended upstream?

lib/Extension/CoreExtension.php
```diff
@@ -344,7 +344,10 @@ private function registerProgressLoggers(Container $container)
         $container->register('progress_logger.registry', function (Container $container) {
             $registry = new LoggerRegistry();
 
-            foreach ($container->getServiceIdsForTag('progress_logger') as $serviceId => $attributes) {
+            /** @var array $serviceIdsForTag */
+            $serviceIdsForTag = $container->getServiceIdsForTag('progress_logger');
```
localheinz: Perhaps this should be amended upstream?

localheinz:
Hmm, tests are failing. Not sure right now, will take a look later.

lib/Benchmark/Remote/Launcher.php
dantleech: Yeah, we could make the return type `?string` ?

lib/Extension/CoreExtension.php
dantleech: hmm, it should already have a docblock return type of `array` at least :thinking: but I wouldn't worry about it for this PR.

phpstan.neon
dantleech: yeah, it will only enter that branch if there is at least one reject.

tests/Unit/Console/Command/LogCommandTest.php
localheinz: You mean, move this to `PhpBench\Console\Application`?

dantleech: I guess the question is why did we change this? Did it recently become part of the command class or? Can we keep it as a method call on the `application` ?

localheinz: The method was removed in `symfony/console`, that's why we either need to stop using it, or implement it ourselves.

lib/Benchmark/Remote/Launcher.php
localheinz: Hmm, just noticed it's `symfony/process`, so here we go: https://github.com/symfony/symfony/pull/29702.

tests/Unit/Console/Command/LogCommandTest.php
dantleech: It's replaced by the [Terminal](https://symfony.com/blog/new-in-symfony-3-2-console-improvements-part-2#introduced-a-new-terminal-class) class.

But if we can replace with env vars, that's probably the best solution :+1:

dantleech:
Awesome, thanks alot!

localheinz:
Thank you, @dantleech!

PR #567, opened 2018-12-24
Enhancement: Normalize composer.json

localheinz:
This PR

* [x] normalizes `composer.json`

💁‍♂️ I ran

```
$ composer global require localheinz/composer-normalize:^1.0.0
```

followed by

```
$ composer normalize
```

For reference, also see

* https://github.com/localheinz/composer-normalize
* https://localheinz.com/blog/2018/01/15/normalizing-composer.json/

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,9 +8,6 @@
             "email": "daniel@dantleech.com"
         }
     ],
-    "config": {
-        "sort-packages": true
-    },
     "require": {
         "php": "^7.1",
         "ext-dom": "*",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -8,36 +8,45 @@
...
+    "suggest": {
+        "ext-curl": "For (web) reports extension",
+        "ext-xdebug": "For XDebug profiling extension."
+    },
+    "config": {
+        "sort-packages": true
+    },
+    "extra": {
+        "branch-alias": {
+            "dev-master": "1.0-dev"
```
dantleech: pretty sure this is actually invalid :flushed:

localheinz: Hmm, running

```
$ composer normalize
```

does not change the semantics, only the format.

Do you want me to change this anyway?

localheinz:
Thank you, @dantleech!

PR #569, opened 2018-12-24
Enhancement: Reference phpunit.xsd as installed with composer

localheinz:
This PR

* [x] references `phpunit.xsd` in `phpunit.xml.dist` as installed with `composer`

💁‍♂️ Helpful with auto-completion!


localheinz:
Thank you, @dantleech!

PR #568, opened 2018-12-24
Enhancement: Run build on PHP 7.3 on Travis

localheinz:
This PR

* [x] adds PHP 7.3 to the Travis CI build matrix

PR #570, opened 2018-12-24
Enhancement: Switch to using build stages on Travis

localheinz:
This PR

* [x] switches to using build stages on Travis CI
* [x] disables Xdebug as early as possible to speed up builds

Blocks #568.

💁‍♂️ The idea is this:

* there is no need to run `php-cs-fixer` or `phpstan` on every *test* build
* there is probably no need to run a static analysis with `phpstan` when `php-cs-fixer` detected code style isses
* there is probably no need to run tests with `phpunit` when `php-cs-fixer` and `phpstan` detected code style or static analysis issues
* it will be easier (and more explicit) to add more builds to the matrix (also see #568)

For reference, see https://docs.travis-ci.com/user/build-stages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,50 +1,97 @@
 language: php
-sudo: false
-
-php:
-    - 7.1
-    - 7.2
 
 cache:
   directories:
     - "$HOME/.composer/cache"
 
-before_install:
-    - composer validate
+stages:
+  - style
+  - stan
+  - test
+  - deploy
 
-before_script:
-    - if [ -n "${SYMFONY_VERSION}" ]; then composer require symfony/symfony:${SYMFONY_VERSION} --no-update; fi;
-    - composer update ${COMPOSER_FLAGS} --prefer-dist
-    - export TZ=Europe/Paris
-
-script:
-    - ./vendor/bin/phpstan analyze --level=1 -c phpstan.neon lib/
-    - ./vendor/bin/phpunit
-    - ./vendor/bin/php-cs-fixer fix --dry-run
-    - php bin/phpbench run --report=env --progress=dots --store
-    - php bin/phpbench run --iterations=1 --revs=1 --config=extensions/dbal/benchmarks/phpbench.json --progress=dots
-
-matrix:
-  fast_finish: true
+jobs:
   include:
-    - php: 7.1
-      env:
-      - COMPOSER_FLAGS="--prefer-lowest"
-    - php: 7.1
-      env: 
-      - SYMFONY_VERSION=^3.2
-      - EXECUTE_DEPLOYMENT=true
+    - stage: Style
 
-before_deploy:
-    - ./.travis/before-deploy.sh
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,50 +1,91 @@
...
   directories:
     - "$HOME/.composer/cache"
 
-before_install:
-    - composer validate
+stages:
+  - style
+  - stan
+  - test
+  - deploy
```
localheinz: This defines the order in which stages will be run.

.travis.yml
```diff
@@ -1,50 +1,91 @@
 language: php
-sudo: false
```
localheinz: See https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration.

.travis.yml
```diff
@@ -1,50 +1,91 @@
...
+      before_install:
+        - composer validate
+
+      install:
+        - composer install
+
+      script:
+        - ./vendor/bin/phpstan analyze --level=1 -c phpstan.neon lib/
+
+    - &TEST
```
localheinz: This defines a template which we will use to add more test jobs.

.travis.yml
```diff
@@ -1,50 +1,91 @@
...
+
+      env: COMPOSER_FLAGS="--prefer-lowest"
+
+    - <<: *TEST
+
+      php: 7.2
+
+    - stage: Deploy
+
+      if: (NOT type IN (pull_request)) AND (branch = master)
```
localheinz: This ensures that the job will be only run when it is a branch build, and we are on `master`.

.travis.yml
```diff
@@ -1,50 +1,91 @@
...
+
+      php: 7.1
+
+      script: skip
```
localheinz: Nothing to do here, unless you prefer to move the script that is run in the `before_deploy` section here, since now we have a dedicated deploy stage.

.travis.yml
```diff
@@ -1,50 +1,107 @@
...
-    github-token: 
-        secure: "cVKcMHHv6MV1HiP78hYPyBx6VIy+DzIv6HS0QM/g3gHnE7EWopu9mQjku2DQtQWAjqVrIPL3ClvZxJH5MjDd4/kB6kWQDGX5O2+CaqzhaJgSpT0htFLbiClqK1uvl2QnHoqwfnhIux230AWDtQfod1g1p60nryo65pLG0HGHqXU="
-    on:
-        branch: master
-        condition: $EXECUTE_DEPLOYMENT == true
+      install:
+        - composer install
+
+      script:
+        - ./vendor/bin/php-cs-fixer fix --diff --dry-run --verbose
```
localheinz: Added `--diff` and `--verbose` options so we can see progress and diff.

dantleech:
:+1: looks great

localheinz:
Thank you, @dantleech, glad you like it!

PR #572, opened 2018-12-24
Enhancement: Enable php_unit_set_up_tear_down_visibility fixer

localheinz:
This PR

* [x] enables the `php_unit_set_up_tear_down_visibility` fixer
* [x] runs `php-cs-fixer`

❗️ Possibly blocked by #571.

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.13.1#usage:

>**php_unit_set_up_tear_down_visibility**
>
>Changes the visibility of the `setUp()` and `tearDown()` functions of PHPUnit to `protected`, to match the PHPUnit TestCase.
>
>*Risky rule: this fixer may change functions named ``setUp()`` or ``tearDown()`` outside of PHPUnit tests, when a class is wrongly seen as a PHPUnit test.*

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -9,11 +9,13 @@ $finder = PhpCsFixer\Finder::create()
 ;
 
 return PhpCsFixer\Config::create()
+    ->setRiskyAllowed(true)
     ->setRules([
         '@PSR2' => true,
         'binary_operator_spaces' => [ 'align_double_arrow' => false ],
         'concat_space' => false,
         'method_argument_space' => false,
+        'php_unit_set_up_tear_down_visibility' => true,
         'phpdoc_align' => [],
         'phpdoc_indent' => false,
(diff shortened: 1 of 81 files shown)


.php_cs.dist
```diff
@@ -9,11 +9,13 @@ $finder = PhpCsFixer\Finder::create()
 ;
 
 return PhpCsFixer\Config::create()
+    ->setRiskyAllowed(true)
```
localheinz: Need to allow risky fixers as the detection of PHPUnit test classes could fail, and we could end up fixing the visibility of `setUp()` and `tearDown()` methods in unrelated classes.

dantleech:
:+1: something I need to start doing on all my projects :)

localheinz:
Thank you, @dantleech!

Issue #537, opened 2018-05-14
Remove styleci, replace with phpcsfixer


dantleech:
Use `phpcsfixer` in travis, remove StyleCI

dantleech:
Done!

PR #573, opened 2018-12-24
Fix: Remove apparently unused configuration for StyleCI

localheinz:
This PR

* [x] removes an apparently unused configuration for StyleCI

Follows #537.


dantleech:
Yep, I removed this sometime ago, not sure why the config was left behind :+1:

localheinz:
Thank you, @dantleech!

PR #579, opened 2018-12-27
Enhancement: Enable phpdoc_separation fixer

localheinz:
This PR

* [x] enables the `phpdoc_separation` fixer

💁‍♂️ For reference, see https://github.com/friendsofphp/php-cs-fixer/tree/v2.13.1#usage:

>**phpdoc_separation** [`@Symfony`]
>
>Annotations in PHPDoc should be grouped together so that annotations of the same type immediately follow each other, and annotations of a different type are separated by a single blank line.

❗️ This code style is basically in use already, but enabling the fixer ensures that it stays that way.


dantleech:
:+1: thanks :)

localheinz:
Thank you, @dantleech!

PR #580, opened 2018-12-27
Enhancement: Enable and configure blank_line_before_statement fixer

localheinz:
This PR

* [x] enables and configures the `blank_line_before_statement` fixer
* [x] runs `php-cs-fixer`

💁‍♂️ For reference, see https://github.com/friendsofphp/php-cs-fixer/tree/v2.13.1#usage:

>**blank_line_before_statement** [`@Symfony`]
>
>An empty line feed must precede any configured statement.
>
>Configuration options:
>
>* `statements` (a subset of `['break', 'case', 'continue', 'declare', 'default', 'die', 'do', 'exit', 'for', 'foreach', 'goto', 'if', 'include', 'include_once', 'require', 'require_once', 'return', 'switch', 'throw', 'try', 'while', 'yield']`): list of statements which must be preceded by an empty line; defaults to `['break', 'continue', 'declare', 'return', 'throw', 'try']`


dantleech:
Less sure about this one, it might be a matter of taste rather than being objectively easier to read. But won't know without trying it :)

localheinz:
Thank you, @dantleech!

PR #581, opened 2018-12-27
Enhancement: Explicitly test against lowest and highest versions of dependencies

localheinz:
This PR

* [x] checks in `composer.lock` and runs tests against lowest, locked, and highest versions of dependencies

### Before

See https://travis-ci.org/phpbench/phpbench/builds/472739856:

screen shot 2018-12-28 at 00 45 47

### After

See https://travis-ci.org/phpbench/phpbench/builds/473001742:

screen shot 2018-12-29 at 12 53 03

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,6 +48,8 @@ jobs:
 
       php: 7.1
 
+      env: WITH_LOWEST=true
+
       before_install:
         - source .travis/xdebug.sh
         - xdebug-disable
@@ -55,8 +57,8 @@ jobs:
 
       install:
         - if [[ "$TRAVIS_PHP_VERSION" == "7.3" ]]; then composer remove --dev friendsofphp/php-cs-fixer; fi
-        - if [ -n "${SYMFONY_VERSION}" ]; then composer require symfony/symfony:${SYMFONY_VERSION} --no-update; fi;
-        - composer update ${COMPOSER_FLAGS} --prefer-dist
+        - if [[ "$WITH_LOWEST" == "true" ]]; then composer update --prefer-lowest; fi
+        - if [[ "$WITH_HIGHEST" == "true" ]]; then composer update; fi
 
       before_script:
         - export TZ=Europe/Paris
@@ -70,21 +72,19 @@ jobs:
 
       php: 7.1
 
-      env: COMPOSER_FLAGS="--prefer-lowest"
-
-    - <<: *TEST
-
-      php: 7.1
-
-      env: SYMFONY_VERSION=^3.2
+      env: WITH_HIGHEST=true
 
     - <<: *TEST
 
       php: 7.2
 
+      env: WITH_HIGHEST=true
+
     - <<: *TEST
 
       php: 7.3
+
+      env: WITH_HIGHEST=true
 
     - stage: Deploy
 


localheinz:
Not sure, do we really need to run a build with `symfony/symfony:^3.2`?

dantleech:
Since it's [not supported](https://symfony.com/roadmap?version=3.2) any more, and there is a 3.4 LTS I guess not?

.travis.yml
```diff
@@ -48,15 +48,18 @@ jobs:
...
         - source .travis/xdebug.sh
         - xdebug-disable
         - composer validate
 
       install:
         - if [[ "$TRAVIS_PHP_VERSION" == "7.3" ]]; then composer remove --dev friendsofphp/php-cs-fixer; fi
-        - if [ -n "${SYMFONY_VERSION}" ]; then composer require symfony/symfony:${SYMFONY_VERSION} --no-update; fi;
-        - composer update ${COMPOSER_FLAGS} --prefer-dist
+        - if [ -n "${SYMFONY_VERSION}" ]; then composer require symfony/symfony:${SYMFONY_VERSION}; fi;
+        - if [[ "$WITH_LOWEST" == "true" ]]; then composer update --prefer-lowest; fi
```
localheinz: Perhaps we should also check in `composer.lock` and run builds against locked versions of dependencies?

```diff
diff --git a/.gitignore b/.gitignore
index 55cecff..8fd2c28 100644
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-composer.lock
 vendor
 docs/build
 .phpbench.sqlite
diff --git a/.travis.yml b/.travis.yml
index bdbf671..68ba9d3 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,6 +59,7 @@ jobs:
         - if [[ "$TRAVIS_PHP_VERSION" == "7.3" ]]; then composer remove --dev friendsofphp/php-cs-fixer; fi
         - if [ -n "${SYMFONY_VERSION}" ]; then composer require symfony/symfony:${SYMFONY_VERSION}; fi;
         - if [[ "$WITH_LOWEST" == "true" ]]; then composer update --prefer-lowest; fi
+        - if [[ "$WITH_LOCKED" == "true" ]]; then composer install; fi
         - if [[ "$WITH_HIGHEST" == "true" ]]; then composer update; fi

       before_script:
@@ -69,6 +70,12 @@ jobs:
         - php bin/phpbench run --report=env --progress=dots --store
         - php bin/phpbench run --iterations=1 --revs=1 --config=extensions/dbal/benchmarks/phpbench.json --progress=dots

+    - <<: *TEST
+
+      php: 7.1
+
+      env: WITH_LOCKED=true
+
     - <<: *TEST

       php: 7.1
@@ -87,6 +94,12 @@ jobs:

       env: WITH_LOWEST=true

+    - <<: *TEST
+
+      php: 7.2
+
+      env: WITH_LOCKED=true
+
     - <<: *TEST

       php: 7.2
@@ -99,6 +112,12 @@ jobs:

       env: WITH_LOWEST=true

+    - <<: *TEST
+
+      php: 7.3
+
+      env: WITH_LOCKED=true
+
     - <<: *TEST

       php: 7.3
```

dantleech: hmm, not sure what you mean, we don't commit the `composer.lock`?

dantleech: I would think we should also try and balance the build time against the version coverage. Previously we just had one PHP version to test the lowest dependencies (IIRC) for this reason. So would be happy to keep the matrix to a minimum and just f.e. test the lowest deps against the lowest PHP version.

Is there any difference between `composer install` and `composer update` as far as the build is concerned?

localheinz: There will only be a difference between

```
$ composer install
```

and 

``` 
$ composer update
```

when `composer.lock` is not checked in, as far as I know.

dantleech: oh sorry, I missed the title of the PR :man_facepalming: 

For libraries (or `require` included tools), I never check-in `composer.lock`. What is the reason for doing so here?

localheinz: Hmm, I think it's a good idea to ensure that the build passes at least against the set of dependencies that were used in development, but I guess we don't have to.

dantleech: but that set of dependencies is arbitrary - there are innumerable combinations of dependencies that will be used in the wild. I don't _think_ there is any particular reason to assume that the state of the application used in development is special in any particular way. The Phar is also built  based on the lastest dependencies.

So, would currently prefer to keep the `composer.lock` unversioned :)

localheinz: Done!

dantleech:
Awesome, thanks :+1:

localheinz:
Thank you, @dantleech!

PR #586, opened 2019-01-01
Enhancement: Raise phpstan level from 4 to 5

localheinz:
This PR

* [x] raises the `phpstan` level from `4` to `5`
* [x] fixes issues detected via static analysis
* [x] ignores errors


dantleech:
Thanks :+1: I think we could fix some of those warnings (e.g. the Registry ContainerInterface one), but this is a good start.

localheinz:
Thank you, @dantleech!

PR #590, opened 2019-02-09
[WIP] Added support for hrtime

marc-mabe:
* added `HrtimeExecutor`
* removed use of microtime template from `BaseExecutor`
* moved template argument from `Launcher::launch` to `Payload::launch`

Funnily I have no idea right now how to use this new executor or how to switch the default executor from `microtime` to `hrtime` if supported.

Also the following executors don't use the required `$template`-argument of `Payload::launch` right now. I couldn't find where they set the template.

Ideas?


lib/Benchmark/Executor/HrtimeExecutor.php
```diff
@@ -0,0 +1,61 @@
...
+        $result = $payload->launch(__DIR__ . '/template/hrtime.template');
+
+        if (isset($result['buffer']) && $result['buffer']) {
+            throw new \RuntimeException(sprintf(
+                'Benchmark made some noise: %s',
+                $result['buffer']
+            ));
+        }
+
+        $iteration->setResult(new TimeResult($result['time']));
```
dantleech: Is `time` still microseconds? I wonder if we should either multiply the HR time to microseconds or specify the unit here (potentially even a new `Result` class).

lib/Benchmark/Executor/template/microtime.template
```diff
@@ -79,7 +79,7 @@ function benchmark($benchmark, $parameters)
         $endTime = microtime(true);
     }
 
-    return ($endTime * 1000000) - ($startTime * 1000000);
+    return ($endTime - $startTime) * 1000000;
```
dantleech: :thinking: can't imagine why it was this way initially

dantleech: could also be `1E6` I guess

lib/Benchmark/Executor/HrtimeExecutor.php
marc-mabe: hrtime returns nanoseconds currently I calculate the microseconds but would be better to keep full accuracy. Also currently I use numeric result which could loose accuracy on 32bit platforms.An alternative would be to return an array of seconds and nanoseconds.

marc-mabe:
It's weird that the microtime executer doesn't define microtime at all. It's defined in the base executer.
Alternatively we could rename it to just time executer and auto detect which time function to use

dantleech:
> It's weird that the microtime executer doesn't define microtime at all.

Yeah, as said this is far from perfect, renaming to `TimeExecutor` would might be an improvement, but then this class has an implicit dependency on the template (it depends on `time` and `mem` return values from that script, and we are already suggesting to return a different value set (`[seconds, nanoseconds]`).

I'm be tempted to suggest the path of least-resistance - create a new executor, extending the BaseExecutor, add a constructor parameter to define the template.

You can then register it like this: https://github.com/phpbench/phpbench/blob/master/lib/Extension/CoreExtension.php#L164

We can then get the build green and think about refactoring? WDYT?

PR #591, opened 2019-02-12
Use hrtime instead of microtime if possible

marc-mabe:
Simplified version of #590

Auto detect `hrtime()` support by detecting if the function exists directly within the executor template to make sure the auto detection runs on the executed PHP binary.

I didn't changed the internally used microtime unit but the microtime value can be a float now.

Renamed executors as the point here is that either the time measurement or the memory measurement gets the attention - the resolution should just be as high as possible (or would you differ between bytememory and mbytememory executor?):
* `microtime` -> `time`
* `memory_centric_microtime` -> `memory_centric_time`
* The old executor names still work but point the the new executors for BC reasons


dantleech:
Thanks, will try and have a proper look at this in the next week. I'm still not decided if:

- This should be a separate executor.
- We should further add time unit support in the rest of Phpbench (e.g. render nanoseconds in reports / [progress loggers), which would be a separate PR.

And invetigate if:

- Adding nanosecond precision has any practical value (there is a considerable margin of error even for doing nothing).
- and then, if there is a difference between microtime and nanosecond samples -- in which case it would be necessary to separate the two I think.

But this looks promising :)

marc-mabe:
>     * This should be a separate executor.
I think it makes most sense to have just one general time executor as the user wants to measure the time in the most reliable way.

>     * We should further add time unit support in the rest of Phpbench (e.g. render nanoseconds in reports / [progress loggers), which would be a separate PR.
Currently this PR still handles everything as microseconds for simplicity of the PR but I think there are two more things to do:
Calculate generated timings internally in the most accurate way. Means, try to avoid floating point numbers or any other precision loosing calculations as much as possible. This could be done to handle times internally as array or seconds and nanoseconds and use more precise mathematical functions if possible (like gmp, bcmath, ...)

>    * Adding nanosecond precision has any practical value (there is a considerable margin of error even for doing nothing).
>     * and then, if there is a difference between microtime and nanosecond samples -- in which case it would be necessary to separate the two I think.
How the numbers are reported is a totally different question.
For sure there are differences between microtime and nanoseconds but I thing a benchmarking framework should choose the best available timer automatically as this makes the timers a bit less flicker and reduces error margin.

dantleech:
> I think it makes most sense to have just one general time executor as the user wants to measure the time in the most reliable way.

If microseconds and nanoseconds return different results (regardless of precision) then it might be beneficial to explicitly state which sampling method you want to use, it is also beneficial for development, e.g. comparing `microseconds` vs `hrtime`. And beyond that, being able to explicitly use one or the other might be useful in situations where you want to compare an algorithm across versions of PHP which don't support `hrtime`: Being able to state that all measurements are `microtime` would then remove a potential variable.

lib/Benchmark/Executor/template/time.template
```diff
@@ -14,6 +14,32 @@ $bootstrap = '{{ bootstrap }}';
...
+    $timeFunc = 'microtime';
+    $timeDiffFunc = function($startTime, $endTime) {
+        list($startMicro, $startSec) = explode(' ', $startTime);
+        list($endMicro, $endSec)     = explode(' ', $endTime);
+        $startSec   = (int)$startSec;
+        $endSec     = (int)$endSec;
+        $startMicro = (int)substr($startMicro, 2, 6);
+        $endMicro   = (int)substr($endMicro, 2, 6);
+        $diffSec    = $endSec - $startSec;
+        $diffMicro  = $endMicro - $startMicro;
```
dantleech: NB: Would prefer not to use whitespace alignment

lib/Benchmark/Executor/template/time.template
```diff
@@ -14,6 +14,32 @@ $bootstrap = '{{ bootstrap }}';
...
+    $timeDiffFunc = function($startTime, $endTime) {
+        list($startSec, $startNano) = $startTime;
+        list($endSec, $endNano)     = $endTime;
+        $diffSec  = $endSec - $startSec;
+        $diffNano = $endNano - $startNano;
+        return ($diffSec * 1000000) + ($diffNano / 1000);
+    };
+} else {
+    // microtime() generates the following format: "0.MMMMMM00 SSSSSSSSSSS"
+    $timeFunc = 'microtime';
```
dantleech: Sorry if I sound ignorant, but why do you change this? Do we lose precision with `microtime(true) * 1E6` ?

dantleech:
(updated comment above)

marc-mabe:
The only difference in microtime vs hrtime is the accuracy.
As you can see here the returned microtime value as string is based on two integer values and without precision loose where the returned float value looses precision: https://github.com/php/php-src/blob/38c337f22eff21ae4123ce1d4354154ee8c63983/ext/standard/microtime.c#L64

lib/Benchmark/Executor/template/time.template
```diff
@@ -14,6 +14,32 @@ $bootstrap = '{{ bootstrap }}';
...
 
+if (function_exists('hrtime')) {
+    // hrtime() generates the following format: [S, NS]
+    $timeFunc = 'hrtime';
+    $timeDiffFunc = function($startTime, $endTime) {
+        list($startSec, $startNano) = $startTime;
+        list($endSec, $endNano)     = $endTime;
+        $diffSec  = $endSec - $startSec;
+        $diffNano = $endNano - $startNano;
+        return ($diffSec * 1000000) + ($diffNano / 1000);
```
dantleech: why not `hrtime(true)` ?

lib/Benchmark/Executor/template/time.template
marc-mabe: changed - should be fine at this point but math operations could be improved on calculating statistics.

marc-mabe:
@dantleech rebased the PR and updated comments

PR #592, opened 2019-02-25
Clean Garbage Collection after each Revolution

julien-boudry:
Memory maintenance is disabled between each Revolution. For obvious reasons of predictability of performance. 
If it is justified for the execution time. But it leads to an uncontrolled memory leak giving a false idea of it in reports, and a risk of exceeding the memory limit.

This correction forces a manual cleaning between each Revolution, in a way that does not impact the time measurement. The time measurement itself is placed a little more precisely in the code.

**Before Fix:**
_Memory increases in proportion to the number of Revolutions to infinity._

* 6 Revolutions :
image

* 12 Revolutions:
image

**After fix (look memory):**

image

### DIFF
--- a/lib/Benchmark/Executor/template/microtime.template
+++ b/lib/Benchmark/Executor/template/microtime.template
@@ -1,8 +1,5 @@
 <?php
 
-// disable garbage collection
-cleanGarbageCollectionAndDisable();
-
 // repress any output from the user scripts
 ob_start();
 
@@ -34,6 +31,9 @@ if ($warmup) {
         $benchmark->{{ subject }}($parameters);
     }
 }
+
+// disable garbage collection
+cleanGarbageCollectionAndDisable();
 
 $time = benchmark($benchmark, $parameters);
 
--- a/lib/Benchmark/Executor/template/microtime.template
+++ b/lib/Benchmark/Executor/template/microtime.template
@@ -1,7 +1,7 @@
 <?php
 
 // disable garbage collection
-gc_disable();
+cleanGarbageCollectionAndDisable();
 
 // repress any output from the user scripts
 ob_start();
@@ -56,30 +56,47 @@ echo serialize([
     'buffer' => $buffer,
 ]);
 
-function benchmark($benchmark, $parameters)
+function benchmark($benchmark, $parameters) : float
 {
+    $time = 0.00;
+
     // run the benchmarks: note that passing arguments to the method slightly increases
     // the calltime, so we explicitly do one thing or the other depending on if parameters
     // are provided.
     if ($parameters) {
-        $startTime = microtime(true);
-
         for ($i = 0; $i < {{ revolutions }}; $i++) {
+            $startTime = microtime(true);
+            
             $benchmark->{{ subject }}($parameters);
-        }
 
-        $endTime = microtime(true);
+            $endTime =  microtime(true);
+
+            $time += ($endTime * 1000000) - ($startTime * 1000000);
+
+            cleanGarbageCollectionAndDisable();
+        }
     } else {
-        $startTime = microtime(true);
-
         for ($i = 0; $i < {{ revolutions }}; $i++) {
-            $benchmark->{{ subject }}();
-        }
+            $startTime = microtime(true);
 
-        $endTime = microtime(true);
+            $benchmark->{{ subject }}();
+
+            $endTime =  microtime(true);
+
+            $time += ($endTime * 1000000) - ($startTime * 1000000);
+
+            cleanGarbageCollectionAndDisable();
+        }
     }
 
-    return ($endTime * 1000000) - ($startTime * 1000000);
+    return $time;
+}
+
+function cleanGarbageCollectionAndDisable () : void
+{
+    gc_enable();
+    gc_collect_cycles();
+    gc_disable();
 }
 
 exit(0);


dantleech:
Looks good :+1:

lib/Benchmark/Executor/template/microtime.template
```diff
@@ -56,30 +56,47 @@ echo serialize([
...
         for ($i = 0; $i < {{ revolutions }}; $i++) {
+            $startTime = microtime(true);
+            
             $benchmark->{{ subject }}($parameters);
-        }
 
-        $endTime = microtime(true);
-    } else {
-        $startTime = microtime(true);
+            $endTime =  microtime(true);
```
dantleech: is there a reason for putting the time measurement in the loop?

I think the idea here is that the SUT might not even take a micro-second, so by putting it outside the loop we are in a better position to judge the average, having it in the loop means we have a larger risk of an inaccurate reading (also `microtime` is not costless).

dantleech: Ah, I see it's fundamental to this approach. I'm not sure this is the best thing to do, would need to do some comparisons to be sure.

dantleech: It might make more sense to have another executor (and another template) which is optimised for memory.

julien-boudry: Hi,

Yes, I don't see how to simply manage it out of the loop for this approach. However, I don't believe in the risks mentioned, it seems unlikely to me (or it's that the measurements are empty!).

In the current case, the indicated memory already does not correspond to any reality, because in real life the Garbage Collection is activated. The behavior of disabling it is only one way to reduce uncontrollable hazards during the time of a measurement of execution time.
On the other hand, not compensating for this defect by cleaning between measures seems to me to form an aggravation of this defect, whereas there is no longer any pragmatic use in doing so.
Moreover, in the case of several Revolutions, the Memory Peak no longer has any connection with reality, giving delirious scores and therefore has no interest.

In addition, for testing an application like mine: likely to manipulate a lot of data. Not applying this correction would form a default behaviour that would tend to greatly exaggerate its memory consumption, and inexorably tend towards an execution error due to lack of memory.

To summarize: the current behaviour is misleading (on the memory used) and tends to run errors in some cases. The proposed correction does not disturb the measurement of time (except in cases where I would not have understood?) which remains constant, while proposing a more realistic default behaviour in terms of memory and limiting the risk of execution error due to lack of memory.

Another point on the other hand: I put the function cleanGarbageCollectionAndDisable here because of lack of idea. But perhaps it would be more relevant elsewhere.

dantleech: > The proposed correction does not disturb the measurement of time

I think the effect on time would have on "micro" benchmarks, i.e. benchmarks that are best measured in microseconds rather than milliseconds (e.g. benchmarking a function call, or an algorithm). I would have to test to be sure, but I suspsect this will have an effect.

If it does, it might make more sense to have "memory-centric" and "time-centric" executors.

julien-boudry:
I just edit my original pull request to add a bugfix.

For some mysterious reason, on windows x64 PHP 7.3.2: The original commit cause randomly this bug:
image


I don't understand why, because my bench are 100% reproducible. This bug appear only ~20% of time and not for each bench methods.
If I put the first cleanGarbageCollectionAndDisable() later, this bug not happen.

dantleech:
```
memory executor:
+--------------+------------+-----+------+-----+------------+----------+----------+----------+----------+----------+--------+-------+
| benchmark    | subject    | set | revs | its | mem_peak   | best     | mean     | mode     | worst    | stdev    | rstdev | diff  |
+--------------+------------+-----+------+-----+------------+----------+----------+----------+----------+----------+--------+-------+
| HashingBench | benchAlgos | 0   | 1000 | 10  | 1,214,400b | 4.6550μs | 4.8411μs | 4.8454μs | 5.1160μs | 0.1262μs | 2.61%  | 1.00x |
| HashingBench | benchAlgos | 1   | 1000 | 10  | 1,214,400b | 5.0570μs | 5.4107μs | 5.2402μs | 6.5920μs | 0.4341μs | 8.02%  | 1.12x |
+--------------+------------+-----+------+-----+------------+----------+----------+----------+----------+----------+--------+-------+

microtime exexutor:
+--------------+------------+-----+------+-----+------------+----------+----------+----------+----------+----------+--------+-------+
| benchmark    | subject    | set | revs | its | mem_peak   | best     | mean     | mode     | worst    | stdev    | rstdev | diff  |
+--------------+------------+-----+------+-----+------------+----------+----------+----------+----------+----------+--------+-------+
| HashingBench | benchAlgos | 0   | 1000 | 10  | 1,212,248b | 3.1570μs | 3.6896μs | 3.3601μs | 5.5700μs | 0.7323μs | 19.85% | 1.00x |
| HashingBench | benchAlgos | 1   | 1000 | 10  | 1,212,248b | 3.5530μs | 4.5041μs | 4.0562μs | 5.5010μs | 0.7002μs | 15.55% | 1.22x |
+--------------+------------+-----+------+-----+------------+----------+----------+----------+----------+----------+--------+-------+


as you can see the memory executor exagerates the time taken at this scale.

I created #594 which adds your script as a separate executor, let me know what you think.

PR #593, opened 2019-03-01
Segregate Executor

dantleech:
This PR segregates and then removes ExecutorInterface.

The current ExecutorInterface has too many responsiblities. By segregating the interface it should be possible to easily introduce new executors while not duplicating other functionality.

- Removes the `ExecutorInterface`
- Introduces
  - `BechmarkExecutorInterface`: The executor which performs the measurement)
  - `MethodExecutorInterface`: Implement to enable an executor to run methods such as before class, after class, etc. *optional*
  - `HealthCheckInterface`: Implement to determine if an executor is capable of running *optional*.
- Applies some other refactorings.
- Introduces `TemplateExecutor` which can be used to create new benchmark templates.
- Introduces `CompositeExecutor` which can combine an benchmark executor, method exeutor and health check.

cc @marc-mabe @julien-boudry This refactoring should make it easier to introduce new executors.


extensions/xdebug/tests/Unit/Executor/TraceExecutorTest.php
```diff
@@ -33,44 +32,93 @@
...
+     * @var ObjectProphecy
+     */
+    private $parameterSet;
+
     protected function setUp()
     {
-        $this->launcher = $this->prophesize(Launcher::class);
+        $this->templateExecutor = $this->prophesize(TemplateExecutor::class);
         $this->filesystem = $this->prophesize(Filesystem::class);
         $this->converter = $this->prophesize(TraceToXmlConverter::class);
```
dantleech: this should be a functional test (we shouldn't be creating this many mocks)

lib/Benchmark/Runner.php
```diff
@@ -325,4 +326,30 @@ public function runIteration(ExecutorInterface $executor, Config $executorConfig
...
+        if (!$executor instanceof MethodExecutorInterface) {
+            return;
+        }
+
+        if (!$benchmarkMetadata->getAfterClassMethods()) {
+            return;
+        }
+
+        $executor->executeMethods($benchmarkMetadata, $benchmarkMetadata->getAfterClassMethods());
+    }
```
dantleech: could probably extract this stuff to a class

lib/Executor/Benchmark/TemplateExecutor.php
```diff
@@ -0,0 +1,100 @@
...
+use PhpBench\Model\Result\TimeResult;
+use PhpBench\Registry\Config;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * This class generates a benchmarking script and places it in the systems
+ * temp. directory and then executes it. The generated script then returns the
+ * time taken to execute the benchmark and the memory consumed.
+ */
+class TemplateExecutor implements BenchmarkExecutorInterface
```
dantleech: remove the doc comment (or at least make it more breif)

lib/Executor/ExecutorInterface.php
```diff
@@ -0,0 +1,21 @@
...
+ *
+ */
+
+namespace PhpBench\Executor;
+
+use PhpBench\Executor\BenchmarkExecutorInterface;
+use PhpBench\Executor\HealthCheckInterface;
+use PhpBench\Executor\MethodExecutorInterface;
+
+interface ExecutorInterface extends BenchmarkExecutorInterface, HealthCheckInterface, MethodExecutorInterface
```
dantleech: this could probably be removed now...

PR #594, opened 2019-03-02
Memory-centric executor

dantleech:
Introduces a new `memory_centric_microtime` executor which does a better job of measuring memory usage.


dantleech:
cc @julien-boudry

julien-boudry:
Look good. How to use it ? Is it in master ?

julien-boudry:
cc @dantleech Can we merge ?

dantleech:
Yes :)

PR #600, opened 2019-05-02
Update dbal.rst

### LINKED ISSUE #601: Documentation is slightly wrong considering the DBAL configuration
The docs give an example on how to configure DBAL for phpbench. The `extensions` array in the config file is not an array but an object in the given example, which is wrong JSON syntax.

Also, just to make clear where to put these configs, it would be nice to add the info that this has to be put into the `phpbench.json`. Which might seem obvious, but just to be extra clear, I think it would be nice. Also, other examples in the docs do have this info.

Christoph-Harms:
Fix wrong JSON syntax and make clear where to put the JSON.

closes #601

### DIFF
--- a/docs/extensions/dbal.rst
+++ b/docs/extensions/dbal.rst
@@ -24,15 +24,16 @@ package installed, install it with composer::
 
     $ composer require --dev "doctrine/dbal"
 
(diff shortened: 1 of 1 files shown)


dantleech:
Thanks :)

PR #602, opened 2019-08-31
Baseline.php : Fix PHP 7.4 notice on fgets()

julien-boudry:
Fix new PHP 7.4 notice on Baseline.php leading to the crash.

image

https://bugs.php.net/bug.php?id=78482

### DIFF
--- a/lib/Benchmark/Baseline/Baselines.php
+++ b/lib/Benchmark/Baseline/Baselines.php
@@ -54,7 +54,7 @@ class Baselines
 
         fclose($handle);
 
-        $handle = fopen($tempName, 'w');
+        $handle = fopen($tempName, 'r');
 
         $line = true;
 


dantleech:
Thanks, pretty surprised that worked at all :scream_cat:

dantleech:
Tagged 0.16.10

julien-boudry:
Thanks @dantleech 
You can also add Travis test just like that:
image

It's will probably become only '7.4' with the first RC in a few days.

Other thing: The new changelog and documentation not say a word about the new memory-centric executor.

ps: it's julien-b**ou**dry and not bodry (changelog)

dantleech:
Oops, I updated the changelog. I didn't include the new executor -- tbh I'm not really sure what the limitations of it are, the tradeoffs, how accurate it is etc. so for now it's an undocumented feature.

PR #605, opened 2019-09-10
Fix tag validation to allow numbers

rybakit:


tests/Unit/Model/TagTest.php
```diff
@@ -18,9 +18,31 @@
...
+    /**
+     * @doesNotPerformAssertions
+     * @testWith
+     * ["foobar"]
+     * ["FooBAR"]
+     * ["42"]
+     * ["foo42"]
+     * ["foo_42"]
+     */
+    public function testValidTagValue(string $tag)
```
dantleech: Nice, didn't know about these annotations, however in other test cases we use `@dataProvider` and `$this->expectException(invalidArgumentException::class)`, e.g.

```
/**
 * @dataProvider provideValidTag
 */
public function testValidTag(string $tag)
{
   //...
}

public function provideValidTag()
{
    yield [
        'foobar'
    ];
}
```

Can we change it to be consistent?

rybakit: Changed: https://github.com/phpbench/phpbench/pull/605/commits/dc8cef0a18331414f130ba815bc56af359a78b41.

PR #611, opened 2019-11-18
Revert "Symfony 5"

dantleech:
@julien-boudry reverting this for now as there are breaking BC breaks (e.g. process requires array not string). I assumed wrongly that the CI would test v5, but as it's still RC it wasn't picked up (and php-cs-fixer requires 4x)


julien-boudry:
Hi Dan,

php-cs-fixer will support Symfony 5 in fews days. Probably before Symfony final release : https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4608

PR #612, opened 2019-11-18
Symfony 5 Support

dantleech:
There are BC breaks in Symfony 5 which will need to be addressed (e.g. process requires an array argument not a string argument).

cc @julien-boudry


dantleech:
See #617

PR #614, opened 2019-11-21
Timeout

### LINKED ISSUE #613: Add a configurable timeout per benchmark
I'm missing a configurable value where I could set a timeout value after which a revolution or iteration is deemed as failed, and is skipped _instantly_.

`@Assert` exists, but that only runs after a benchmark method has run its course.

Could there be a `@Timeout` that skips benchmarks where either a revolution or an iteration takes too long?

dantleech:
Introduces an annotation for specifying the process timeout for iterations cc @rask 

Fixes #613

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,6 +37,13 @@
         "phpunit/phpunit": "^8.5",
         "phpspec/prophecy": "^1.8"
     },
+    "scripts": {
+        "integrate": [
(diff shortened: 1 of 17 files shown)


rask:
Awesome! Looks like this did not require as much code as I thought it would, which is good.

PR #623, opened 2020-02-02
reduced time/mem centric executor overhead

marc-mabe:
* parse `parameters` directly into parameter positions -> no variable and parameter condition needed
* removed function `backmark` and instead execute the banchmark in place -> this reduces the function call stack by one

PS: I would like to introduce a condition for hrtime (see #591) to not need to call time function dynamically which would increase time function call overhead.

### DIFF
--- a/lib/Executor/Benchmark/TemplateExecutor.php
+++ b/lib/Executor/Benchmark/TemplateExecutor.php
@@ -45,6 +45,8 @@ class TemplateExecutor implements BenchmarkExecutorInterface
 
     public function execute(SubjectMetadata $subjectMetadata, Iteration $iteration, Config $config): void
     {
+        $parameterSet = $iteration->getVariant()->getParameterSet();
+
         $tokens = [
             'class' => $subjectMetadata->getBenchmark()->getClass(),
             'file' => $subjectMetadata->getBenchmark()->getPath(),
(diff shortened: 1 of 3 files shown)


lib/Executor/Benchmark/TemplateExecutor.php
```diff
@@ -45,14 +45,16 @@ public function __construct(Launcher $launcher, string $templatePath)
...
+
         $tokens = [
             'class' => $subjectMetadata->getBenchmark()->getClass(),
             'file' => $subjectMetadata->getBenchmark()->getPath(),
             'subject' => $subjectMetadata->getName(),
             'revolutions' => $iteration->getVariant()->getRevolutions(),
             'beforeMethods' => var_export($subjectMetadata->getBeforeMethods(), true),
             'afterMethods' => var_export($subjectMetadata->getAfterMethods(), true),
-            'parameters' => var_export($iteration->getVariant()->getParameterSet()->getArrayCopy(), true),
+            'parameters' => $parameterSet->count() ? var_export($parameterSet->getArrayCopy(), true) : '',
```
dantleech: could revert this I guess?

dantleech:
I'll update the changelog and fix the division by zero in master if it happens again.

PR #632, opened 2020-02-17
Improve extendablity of TemplateExecutor

jaapio:


dantleech:
Seems like there is a cs err-  can be fixed with `./vendor/bin/php-cs-fixer fix` I guess

jaapio:
Fix, should be ready to merge, if you want to :-)

dantleech:
Cheers :)

PR #645, opened 2020-08-05
Remove the reference to Athletic from the docs

MidnightDesign:
Athletic is no longer maintained and shouldn't be considered an alternative.


dantleech:
True, thanks

PR #654, opened 2020-08-26
Refactor baseline to facilitate assertions

dantleech:
This PR refactors the existing baseline logic:

- Baseline is included at the "model" stage, before reports.
- Absolute baseline values are available in 
- Facilicate assertions relative to baselines 
- Removes dedicated `aggregate_baseline` report -- baselines work by default if a suite is included when running.


lib/Console/Command/RunCommand.php
```diff
@@ -131,6 +131,10 @@ public function execute(InputInterface $input, OutputInterface $output)
 
         $suite = $this->runnerHandler->runFromInput($input, $output, $config);
 
+        if ($input->getOption('uuid') || $input->getOption('file')) {
+            $suite->mergeBaselines($this->suiteCollectionHandler->suiteCollectionFromInput($input));
+        }
```
dantleech, line 136: We can refactor to include the baselines in the model _before_ running the benchmarks. This would allow the assertions to take the baselines into account.

lib/Model/Suite.php
```diff
@@ -250,4 +258,41 @@ public function generateUuid()
...
+                    $variant->getParameterSet()->getName()
+                );
+
+                if (!$baselineVariant) {
+                    continue;
+                }
+
+                $variant->attachBaseline($baselineVariant);
+            }
+        }
```
dantleech, line 281: Multiple suites can be used to provide baselines, in case of multiple matching variants, the last matching variant wins.

PR #655, opened 2020-08-27
Punctuation: replace dot with comma

a-ast:
A comma would make more sense in this sentence. 
Forget it if that was intended.


dantleech:
Thanks :)

PR #656, opened 2020-08-28
Assertion Refactoring: Use DSL and allow assertions on baseline

dantleech:
This PR refactores the assertion feature:

- Use a DSL e.g. `@Assert("variant.mean < 100 microseconds +/- 10 microseconds")`
- Allow assertion against baseline if set: `@Assert("variant.mode < baseline.mode +/- 10 milliseconds")`

The DSL is using the Parsica library which requires PHP 7.4 - need to decide if we bump the whole library to 7.4 or backport some or all of Parsica to 7.2/7.3.

Features:

- `@Assert("variant.mode within 10% +/- 5% of baseline.mode`)#
- `phpbench run benchmarks/MyBench.php --assert="variant.mode < 5 microseconds"`

TODO:

- [x] Support all comparison operators
- [x] Format time units in assertions (e.g. `microseconds` => `μs`, `+/-` => `±`)
- [x] Format perentage values
- [x] Support tolerance with comparisons
- [ ] Support logical AND//OR (?)
- [x] Support throughput assertions (e.g. `@Assert("variant.mode throughput > 100 rps`) ??)
- [x] Check `%` cases (e.g. working with tolerance)
- [x] Remove `WITHIN` operator
- [ ] Remove warnings?


lib/Math/Distribution.php
```diff
@@ -187,4 +187,8 @@ public function offsetUnset($offset)
     {
         throw new \BadMethodCallException('Distribution is read-only');
     }
+
+    public function toArray()
+    {
+    }
```
dantleech, line 193: remove

lib/Assertion/ExpressionParser.php
```diff
@@ -0,0 +1,150 @@
...
+use function Verraes\Parsica\collect;
+use function Verraes\Parsica\float;
+use function Verraes\Parsica\integer;
+use Verraes\Parsica\Parser;
+use function Verraes\Parsica\sepBy2;
+use function Verraes\Parsica\string;
+use function Verraes\Parsica\stringI;
+use function Verraes\Parsica\whitespace;
+
+class ExpressionParser
```
dantleech, line 34: Create interface and add cached parser.

tests/Unit/Progress/Logger/VerboseLoggerTest.php
```diff
@@ -110,7 +112,7 @@ public function testLogError()
     {
         $this->variant->hasErrorStack()->willReturn(true);
         $this->variant->getSubject()->willReturn($this->subject->reveal());
-        $this->variant->hasFailed()->willReturn(false);
+        $this->variant->getAssertionResults()->willReturn(new VariantAssertionResults($this->variant->reveal(), []));
```
dantleech, line 115: Should really refactor so that these mocks can be removed.

PR #661, opened 2020-09-06
Config profiles

dantleech:


composer.json
```diff
@@ -34,11 +34,12 @@
         "doctrine/dbal": "^2.4",
         "friendsofphp/php-cs-fixer": "^2.13.1",
```
dantleech, line 35: can remove symfony/de bug

PR #663, opened 2020-10-17
Various: Refactor payload launcher, debug options and documentation

dantleech:
This PR:

- Refactors the payload launcher:
   - Use builder pattern and the `Launcher` now becomes the `Payload` factory
   - Can define the expected results from the payload
- Added Template executor options to specify the script render path and if it should be removed after execution.
- Added `development` documentation


lib/Benchmark/Remote/Payload.php
```diff
@@ -72,46 +72,60 @@ class Payload
...
+
+    /**
+     * @var bool
+     */
+    private $disableIni;
+
+    /**
+     * @var string|null
+     */
+    private $wrapper;
```
dantleech, line 88: dead code?

tests/Util/BenchmarkMetadataBuiler.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+namespace PhpBench\Tests\Util;
+
+use PhpBench\Benchmark\Metadata\BenchmarkMetadata;
+
+class BenchmarkMetadataBuiler
```
dantleech, line 7: these builder classes are probably superfluous, can probably remove them

dantleech:
Remove in favor of #670

PR #696, opened 2020-11-16
Enhancement: Use run command the same way

OskarStark:


docs/quick-start.rst
```diff
@@ -120,7 +120,7 @@ Now you can execute the benchmark as follows:
 
 .. code-block:: bash
 
-   $ vendor/bin/phpbench run tests/Benchmark/TimeConsumerBench.php --report=default
+   $ php vendor/bin/phpbench run tests/Benchmark/TimeConsumerBench.php --report=default
```
dantleech, line 123: why the `php` prefix? `bin/phpbench` already has the shebang?

dantleech:
... and thanks for all the fixes, agree with all of them :+1:

docs/quick-start.rst
OskarStark: Just because the other commands in this quick start have it 🤔 

If you are fine, I would remove the `php` from `php phpbench ...` completely through the whole docs

dantleech: Yes I think that's best if there is no advantage to it

dantleech: The original document was written many years ago...

OskarStark: Can you merge this? I will provide another PR afterwards

PR #710, opened 2020-11-16
Enhancement: Streamline composer.json

OskarStark:


composer.json
```diff
@@ -50,7 +50,7 @@
         ]
     },
     "suggest": {
-        "ext-curl": "For (web) reports extension",
+        "ext-curl": "For (web) reports extension.",
```
dantleech, line 53: this can actually be removed there is no longer a (web) reports extension :eyes:

OskarStark: Done

PR #723, opened 2020-11-29
Do not deploy to GH releases with Travis

dantleech:
cc @Jeckerson we disabled Xdebug during the build previously on Travis -- not sure why (@localheinz ?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,27 +31,3 @@ jobs:
           secure: "cVKcMHHv6MV1HiP78hYPyBx6VIy+DzIv6HS0QM/g3gHnE7EWopu9mQjku2DQtQWAjqVrIPL3ClvZxJH5MjDd4/kB6kWQDGX5O2+CaqzhaJgSpT0htFLbiClqK1uvl2QnHoqwfnhIux230AWDtQfod1g1p60nryo65pLG0HGHqXU="
         on:
           branch: master
-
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -31,27 +31,3 @@ jobs:
...
         on:
           branch: master
-
-    - stage: Deploy to GH Releases
-
-      php: 7.2
-
-      before_install:
-        - source .travis/xdebug.sh
-        - xdebug-disable
```
localheinz, line 41: Yes, that probably wasn't necessary here/

.travis.yml
```diff
@@ -31,27 +31,3 @@ jobs:
...
-
-      script: skip
-
-      before_deploy:
-        - ./.travis/before-deploy-gh-release.sh
```
localheinz, line 46: That is, unless this script would be slowed down by enabling Xdebug.

Jeckerson (approved):

.travis.yml
dantleech: Probably slows it down - but I would imagine not worryingly

dantleech:
Now all we need to do is remove the last Travis task which uploads the Phar to GH-Pages for the self-update functionality:

- Remove the `self-update` feature.
- Make the `self-update` feature use the lastest GH release

PR #728, opened 2020-12-01
Improve error reporting

dantleech:
Show PHPUnit style success indicator when finishing the run:

image

... and removes some of the less interesting information from this summary line (nb. itreations and revs).

image

### DIFF
--- a/lib/Extension/CoreExtension.php
+++ b/lib/Extension/CoreExtension.php
@@ -172,7 +172,7 @@ class CoreExtension implements ExtensionInterface
             }
 
             $output = new ConsoleOutput();
-            $output->getFormatter()->setStyle('greenbg', new OutputFormatterStyle('black', 'green', []));
+            $output->getFormatter()->setStyle('success', new OutputFormatterStyle('black', 'green', []));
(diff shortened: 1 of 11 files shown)


lib/Extension/CoreExtension.php
```diff
@@ -170,9 +171,13 @@ public function load(Container $container): void
                 return new NullOutput();
             }
 
-            return new ConsoleOutput();
+            $output = new ConsoleOutput();
+            $output->getFormatter()->setStyle('greenbg', new OutputFormatterStyle('black', 'green', []));
```
dantleech, line 175: rename to success

lib/Model/ErrorStack.php
```diff
@@ -62,4 +63,12 @@ public function getIterator(): \ArrayIterator
     {
         return new \ArrayIterator($this->errors);
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function count(): int
+    {
+        return count($this->errors);
```
dantleech, line 72: this was showing as 2 instead of 1 in a local test?

PR #737, opened 2020-12-22
refactor: SC2086 and SC2154 error of shellcheck

KEINOS:
Small fixes for shell script static analysis, [Shell Check](https://github.com/koalaman/shellcheck).
- [SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086): Double quote to prevent globbing and word splitting.
- [SC2154](https://github.com/koalaman/shellcheck/wiki/SC2154): Throw error if the referenced variable is  not assigned.


dantleech:
Thanks :+1:

PR #740, opened 2020-12-29
Rename --uuid option to ref and remove `tag:` prefix

### LINKED ISSUE #733: Rename `--uuid` option to `ref`
The `--uuid` is misleading as you can also supply `--uuid=tag:foobar`. `--ref=` makes more sense in context.

It would also to remove the `tag:` prefix. A (PHPBench) UUID is of a fixed length and can be identified easily, and using a tag is the most common use case. 

So for example:

```
phpbench run --tag=foobar
phpbench run --ref=foobar
```

vs

```
phpbench run --tag=foobar
phpbench run --uuid=tag:foobar
```

dantleech:
Fixes #733 

This PR removes the `--uuid` option and replaces it with `--ref`, and removes the `tag:` prefix from references.

The UUID option originally only accepted PHPBench UUIDs, but in practice you rarely pass these, instead specifying a `tag:foobar`.

The `--ref` option now accepts a UUID, tag or special word (e.g. `latest`)


docs/storage.rst
```diff
@@ -106,55 +105,23 @@ would normally specify a UUID, including queries.
...
 .. code-block:: bash
 
-    $ phpbench report --uuid=tag:my_tag --report=aggregate
-
-.. _archive:
-
-Archiving
----------
-
-Archiving provides a way to export and reimport data from and to the
```
dantleech, line 135: This feature was removed

lib/Console/Command/Handler/SuiteCollectionHandler.php
```diff
@@ -13,43 +13,57 @@
...
         $this->xmlDecoder = $xmlDecoder;
         $this->storage = $storage;
-        $this->uuidResolver = $uuidResolver;
+        $this->refResolver = $refResolver;
     }
 
     public static function configure(Command $command): void
     {
-        $command->addOption('uuid', null, InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY, 'Run UUID');
+        $command->addOption('ref', null, InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY, 'Run UUID');
```
dantleech, line 53: Incorreect description

PR #741, opened 2020-12-29
Fix .phar compilation during release

Jeckerson:
Due Github Actions updated PHP version to the latest (8.0) in their images, latest release failed and didn't uploaded `.phar` into it.

Working example - https://github.com/Jeckerson/phpbench/runs/1622156785?check_suite_focus=true


dantleech:
Yes I noticed this, thanks for the fix !

PR #742, opened 2020-12-29
Improve progress output when baseline (reference) is used

dantleech:
Improve the progress output when using a baseline (and make the output more succinct)

- By default only show the mode and relative deviation (the KDE mode has consistently been a better indicator than the mean)
- When baseline is present, show comparison in output
- Will allow custom formatting (e.g. add more fields to the output)

image

### DIFF
--- a/lib/Progress/VariantSummaryFormatter.php
+++ b/lib/Progress/VariantSummaryFormatter.php
@@ -58,8 +58,7 @@ final class VariantSummaryFormatter
         TimeUnit $timeUnit,
         string $format = self::DEFAULT_FORMAT,
         string $baselineFormat = self::BASELINE_FORMAT
-    )
-    {
+    ) {
         $this->format = $format;
         $this->timeUnit = $timeUnit;
         $this->baselineFormat = $baselineFormat;
@@ -134,6 +133,7 @@ final class VariantSummaryFormatter
             if (!$results->count()) {
                 return self::FORMAT_NONE;
             }
+
             if ($results->failures()->count()) {
                 return self::FORMAT_FAILURE;
             }
--- a/lib/Progress/VariantSummaryFormatter.php
+++ b/lib/Progress/VariantSummaryFormatter.php
@@ -3,6 +3,7 @@
 namespace PhpBench\Progress;
 
 use Closure;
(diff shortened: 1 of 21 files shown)


lib/Progress/VariantSummaryFormatter.php
```diff
@@ -0,0 +1,145 @@
...
+            'variant.variance' => self::NOT_APPLICABLE,
+            'baseline.min' => self::NOT_APPLICABLE,
+            'baseline.max' => self::NOT_APPLICABLE,
+            'baseline.mean' => self::NOT_APPLICABLE,
+            'baseline.mode' => self::NOT_APPLICABLE,
+            'baseline.stdev' => self::NOT_APPLICABLE,
+            'baseline.rstdev' => self::NOT_APPLICABLE,
+            'baseline.variance' => self::NOT_APPLICABLE,
+            'percent_difference' => self::NOT_APPLICABLE,
+            'diff_format' => self::FORMAT_NO_CHANGE,
```
dantleech, line 72: rename => `assertion_style`, `outcome_style`

lib/Progress/VariantSummaryFormatter.php
```diff
@@ -0,0 +1,145 @@
...
+     * @var TimeUnit
+     */
+    private $timeUnit;
+
+    /**
+     * @var string
+     */
+    private $baselineFormat;
+
+    public function __construct(TimeUnit $timeUnit, string $format = self::DEFAULT_FORMAT, string $baselineFormat = self::BASELINE_FORMAT)
```
dantleech, line 36: add test here

tests/Unit/Progress/Logger/BlinkenLoggerTest.php
```diff
@@ -161,6 +161,6 @@ public function testIterationEndStats()
         $this->variant->computeStats();
 
         $this->logger->variantEnd($this->variant);
-        $this->assertStringContainsString('RSD/r: 0.00%', $this->output->fetch());
+        $this->assertStringContainsString('±0.00%', $this->output->fetch());
```
dantleech, line 164: this is testing something for which this clsas is no longer responsible, same for other similar cases here.

PR #743, opened 2021-01-10
No warnings, only success within tolerance

dantleech:
"Warning" was the assertion status when a benchmark fails within in a negative way against an assertion.

This meant:

- If a benchmark is _faster_ than a reference benchmark, it is OK
- If it is slower than a reference benchmark within the tolerance zone then is is a WARNING

In reality it can be expected for benchmarks to fail equally on either side, and so "WARNING" is inappropriate: if the result falls in the tolerance zone then it is OK.

This PR removes the WARNING feature completely:

- Benchmarks that fall within the tolerance zone of an assertion are "OK".


lib/Console/AnsiOutput.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+namespace PhpBench\Console;
+
+use Symfony\Component\Console\Output\OutputInterface;
+
+final class AnsiOutput
```
dantleech, line 7: remove

PR #748, opened 2021-01-16
New Assertion/Expression DSL

dantleech:
The existing expression language is quite limited:

- it only allows you to assert against a pre-defined set of values (included pre-calculated ones)
- property values cannot be assigned a time unit.

In addition, this is an **expression** language which can be used in other parts of PHPBench (e.g. progress formatting, reports etc)

This PR adds a new assertion language which allows functions, and _all_ the results for an iteration are available (i.e. also those provided by extensions):

## Before

```
variant.mode = baseline.mode +/- 10%
```
Above the _mode_ (and all other statistics) are eagerly calculated before the asseriion is run, 

## After

```
mode(variant.time.net) = mode(baseline.time.net) +/- 10%
```

We access the result data from the variant directly and lazily evaluate the mode.

## TODO

- [x] Allow converting microseconds to another time unit for display purposes (e.g. `foo.bar as minutes`)
- [x] Support throughput `ops/second`
- [x] Refactor the parser
- [ ] Basic infix arithimetic
- [ ] Ignore or issue warnings on missing values (e.g. memory values not available with local executor)


lib/Assertion/ExpressionEvaluator.php
```diff
@@ -28,42 +30,39 @@ class ExpressionEvaluator
...
     {
+        if ($node instanceof IntegerNode) {
+            return $node->value();
+        }
+
+        if ($node instanceof FloatNode) {
+            return $node->value();
+        }
+
         if ($node instanceof Comparison) {
```
dantleech, line 58: Rebname all the existing nodes to `*Node`

lib/Assertion/ExpressionEvaluatorFactory.php
```diff
@@ -2,15 +2,23 @@
...
+    }
+
     /**
      * @param array<string,mixed> $args
      */
     public function createWithArgs(array $args): ExpressionEvaluator
     {
-        return new ExpressionEvaluator(new NodeMessageFormatter($args), $args);
+        return new ExpressionEvaluator($args, $this->functions);
     }
```
dantleech, line 23: Can refactor to pass `Node $node, array $args` rather than injet the args in the ctor and use a factory.

lib/Assertion/ExpressionParser.php
```diff
@@ -13,188 +13,299 @@
...
-use PhpBench\Util\MemoryUnit;
-use PhpBench\Util\TimeUnit;
+use RuntimeException;
 
 class ExpressionParser
 {
+    /**
+     * @var Node[]
+     */
+    private array $parts = [];
```
dantleech, line 35: not 7,2

lib/Assertion/ExpressionParser.php
```diff
@@ -13,188 +13,299 @@
...
 
+    public function __construct(ExpressionLexer $lexer)
+    {
+        $this->lexer = $lexer;
+    }
+
     public function parse(string $expression): Node
     {
-        $this->lexer = new ExpressionLexer($expression);
+        $this->lexer->setInput($expression);
```
dantleech, line 49: can refactor this class to purposefully build a stream of Nodes which can be popped and shifted

lib/Assertion/Func/MeanFunction.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace PhpBench\Assertion\Func;
+
+use PhpBench\Math\Statistics;
+
+final class MeanFunction
```
dantleech, line 7: add an interface with a `name()` method

dantleech:
Split to #750

PR #758, opened 2021-03-12
String Concatentation and Better Highlighting

dantleech:
- Functions now accept and return AST nodes
- Introduced `con ~ cat` operator
- Support multi-line expressions
- Adds percentage-difference node
- Variant Summary Formatter refactored to use Expressions

TODO:

- [ ] Provide a way to indicate assertion failure / success in the (progress) formatter again?

### DIFF
--- a/lib/Expression/ExpressionLanguage.php
+++ b/lib/Expression/ExpressionLanguage.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace PhpBench\Expression;
+
+use PhpBench\Expression\Ast\Node;
+
+class ExpressionLanguage
+{
+    /**
+     * @var Lexer
+     */
+    private $lexer;
+    /**
+     * @var Parser
+     */
+    private $parser;
+
+    public function __construct(Lexer $lexer, Parser $parser)
+    {
+        $this->lexer = $lexer;
+        $this->parser = $parser;
+    }
+
+    public function parse(string $expression): Node
+    {
+        return $this->parser->parse($this->lexer->lex($expression));
+    }
+}
--- a/lib/Expression/Func/MeanFunction.php
+++ b/lib/Expression/Func/MeanFunction.php
@@ -2,15 +2,13 @@
 
 namespace PhpBench\Expression\Func;
 
+use PhpBench\Expression\Ast\ListNode;
 use PhpBench\Math\Statistics;
 
 final class MeanFunction
 {
-    /**
-     * @param numeric[] $values
-     */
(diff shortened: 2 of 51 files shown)


lib/Expression/Func/MeanFunction.php
```diff
@@ -2,15 +2,15 @@
...
+use PhpBench\Expression\Ast\PhpValueFactory;
 use PhpBench\Math\Statistics;
 
 final class MeanFunction
 {
-    /**
-     * @param numeric[] $values
-     */
-    public function __invoke(array $values): float
+    public function __invoke(ListNode $values): Node
```
dantleech, line 12: PhpValue

lib/Expression/ExpressionLanguage.php
```diff
@@ -0,0 +1,28 @@
...
+
+    public function __construct(Lexer $lexer, Parser $parser)
+    {
+        $this->lexer = $lexer;
+        $this->parser = $parser;
+    }
+
+    public function parse(string $expression): Node
+    {
+        return $this->parser->parse($this->lexer->lex($expression));
```
dantleech, line 26: Avoids the need to pass both the lexer and the parser everwhere

lib/Expression/Parselet/ConcatParselet.php
```diff
@@ -0,0 +1,32 @@
...
+    {
+        $binaryOperator = $tokens->chomp();
+        $right = $parser->parseExpression($tokens);
+
+        return new ConcatNode($left, $right);
+    }
+
+    public function precedence(): int
+    {
+        return Precedence::COMPARISON_EQUALITY;
```
dantleech, line 30: use dedicated percedence

tests/Unit/Progress/VariantSummaryFormatterTest.php
```diff
@@ -2,100 +2,45 @@
...
-                    $variant->addAssertionResult(AssertionResult::tolerated('ok'));
-                    $this->createBaseline($variant);
-
-                    return $variant;
-                },
-                '%result_style%',
-                'result-neutral'
-            ];
-
-        yield 'assertion fail' => [
```
dantleech, line 83: the variant formatter will now not show SUCCESS / FAIL - make this possible?

dantleech: probably by introducing a ternary operator ... or show individual assertion results in the progress logger -- not now.

PR #762, opened 2021-03-13
Format Feature

dantleech:
Allow a custom progress-logger format to be specified via CLI `--format="mean(variant.time.avg) as ms"` or via. annotation/attribute.


lib/Benchmark/Metadata/SubjectMetadata.php
```diff
@@ -332,6 +337,16 @@ public function getAssertions(): array
         return $this->assertions ?: [];
     }
 
+    public function setFormat(string $format): void
+    {
+        $this->format = $format;
+    }
+
+    public function getFormat(): ?string
```
dantleech, line 345: should refactor this to simply use properties.

lib/Benchmark/RunnerConfig.php
```diff
@@ -365,6 +370,19 @@ public function withAssertions(array $assertions = null): self
         return $new;
     }
 
+    public function withFormat(?string $format = null): self
+    {
+        $new = clone $this;
+        $new->format = $format;
```
dantleech, line 376: this too... (why is it immutable ? )

lib/Progress/VariantSummaryFormatter.php
```diff
@@ -80,19 +80,11 @@ public function formatVariant(Variant $variant): string
     {
         $data = $this->paramProvider->provideFor($variant);
 
-        if (!$this->initialized) {
-            $this->initialize();
-        }
+        $this->normalNode = $this->parser->parse($variant->getSubject()->getFormat() ?? $this->format);
+        $this->baselineNode = $this->parser->parse($variant->getSubject()->getFormat() ?? $this->baselineFormat);
```
dantleech, line 84: this is a big performance regression (aside from the pointless assign). The parsing should be cached to avoid needing to initialize things here..

PR #767, opened 2021-03-16
Support true colors (where available)

dantleech:
This PR aims to:

- Add a true color theme and automatically use the Symfony version supports it
- Apply a color gradient to percentage difference (from green to neutral (white) to red)

image


lib/Expression/ColorMap/SolarisedColorMap.php
```diff
@@ -0,0 +1,81 @@
...
+use PhpBench\Expression\Ast\PercentDifferenceNode;
+use PhpBench\Expression\Ast\StringNode;
+use PhpBench\Expression\Ast\ToleratedTrue;
+use PhpBench\Expression\ColorMap;
+use PhpBench\Util\Gradient;
+
+/**
+ * @implements ColorMap<Node>
+ */
+class SolarisedColorMap implements ColorMap
```
dantleech, line 24: Using the Solarized theme doesn't make much sense unless it's applied to all styles in PHPBench.

This can be a 256 color theme, using mostly the standard 8 bit palette with gradients and more subtle colors of  time units etc.

lib/Expression/ColorMap/SolarisedColorMap.php
```diff
@@ -0,0 +1,81 @@
...
+        /** @phpstan-ignore-next-line */
+        return [
+            FunctionNode::class => 'fg=' . self::GREEN,
+            ParenthesisNode::class => 'fg=' . self::RED,
+            PercentDifferenceNode::class => function (Node $node) {
+                assert($node instanceof PercentDifferenceNode);
+                $gradient = array_merge(
+                    Gradient::calculateHex(self::GREEN, self::BASE2, 100),
+                    Gradient::calculateHex(self::BASE2, self::RED, 100)
+                );
```
dantleech, line 60: this should be either pre-calculated or calculated only once.

lib/Extension/ExpressionExtension.php
```diff
@@ -195,7 +198,7 @@ public function load(Container $container): void
         $container->register(HighlightingNodePrinter::class, function (Container $container) {
             return new HighlightingNodePrinter(
                 $container->get(NodePrinters::class),
-                new Standard8ColorMap()
+                new SolarisedColorMap()
```
dantleech, line 201: the theme should be selected based on the Symfony Console version (I think 256 colors only supported in 5)

lib/Extension/ExpressionExtension.php
```diff
@@ -195,7 +197,7 @@ public function load(Container $container): void
         $container->register(HighlightingNodePrinter::class, function (Container $container) {
             return new HighlightingNodePrinter(
                 $container->get(NodePrinters::class),
-                new Standard8ColorMap()
+                class_exists(NullOutputFormatter::class) ? new TrueColorMap() : new Standard8ColorMap()
```
dantleech, line 200: Symfony Console only supports true color in 5.0: This is a class only exists in 5.0 - another option would be to use PackageVersions

lib/Expression/ColorMap/Util/Color.php
```diff
@@ -0,0 +1,69 @@
...
+
+        return new self(...array_map(function (string $hex): int {
+            return (int)hexdec($hex);
+        }, $hexRgb));
+    }
+
+    /**
+     * @return array{int,int,int}
+     */
+    public function toTuple(): array
```
dantleech, line 49: `toRgb`?

PR #771, opened 2021-03-20
Dedicated Unit Value node

dantleech:
- Allow the display unit to be _evaluated_ (for dynamically determining the unit based on the subject's preference)
- Make the `unit` a node - allowing it to be pretty printed.


lib/Expression/NodePrinter/DisplayAsPrinter.php
```diff
@@ -39,18 +41,32 @@ public function print(Printer $printer, Node $node, array $params): ?string
             return null;
         }
 
-        $unit = $node->as();
+        $unitNode = $node->as();
```
dantleech, line 44: This entire class is getting out of control.

Extract a dedicated unit converter and formatter instead of integrating with the shitty TimeUnit and MemoryUnit.

PR #773, opened 2021-03-22
Introduce new unit converter

dantleech:


lib/Assertion/ParameterProvider.php
```diff
@@ -14,6 +15,7 @@ public function provideFor(Variant $variant)
     {
         return (function (array $variantData) use ($variant) {
             return [
+                'subject' => $this->subjectData($variant->getSubject()),
                 'variant' => $variantData,
                 'baseline' => $variant->getBaseline() ? $this->buildVariantData($variant->getBaseline()) : $variantData,
```
dantleech, line 20: this is a bit of a smell - `subject` can have "properties" but `variant` is a list of results, maybe rename to `result`? this would free-up the `variant` key

lib/Assertion/ParameterProvider.php
```diff
@@ -47,4 +49,15 @@ private function buildVariantData(Variant $variant)
...
     }
+
+    /**
+     * @return array<string,mixed>
+     */
+    private function subjectData(Subject $subject): array
+    {
+        return [
+            'time_unit' => $subject->getOutputTimeUnit(),
+            'time_precision' => $subject->getOutputTimePrecision(),
```
dantleech, line 60: ideally I think having tabular data would be better, this would enable the reports, assertions and progress loggers to operate on the same data structure

lib/Expression/NodePrinter/DisplayAsPrinter.php
```diff
@@ -42,60 +39,26 @@ public function print(Printer $printer, Node $node, array $params): ?string
...
+        $unitValue = $unitNode->value();
 
-        if ($unit === self::DEFAULT_MEMORY_UNIT) {
-            return $this->memoryUnit(
-                $value->value(),
-                MemoryUnit::BYTES
+        return (function (UnitValue $value) use ($printer, $unitValue, $params) {
+            return sprintf(
+                '%s%s',
+                number_format($value->value(), floor($value->value()) === $value->value() ? 0 : 3),
```
dantleech, line 58: precision can come from the display-as-node now

dantleech:
Nope

PR #775, opened 2021-03-22
Subject precision time unit

dantleech:
Ensure that subject precision and time are passed to the time unit for formatting.


lib/Expression/NodeEvaluator/ParameterEvaluator.php
```diff
@@ -40,6 +41,10 @@ public function evaluate(Evaluator $evaluator, Node $node, array $params): Node
             return new StringNode($value);
         }
 
+        if (is_null($value)) {
+            return new NullNode();
```
dantleech, line 45: test?

lib/Expression/NodePrinter/DisplayAsPrinter.php
```diff
@@ -128,4 +102,17 @@ private function memoryUnit(float $value, string $unit): string
     {
         return sprintf('%s %s', MemoryUnit::convertTo($value, MemoryUnit::BYTES, $unit), $unit);
     }
+
+    private function resolvePrecision(?Node $node): ?int
+    {
+        if (null === $node) {
```
dantleech, line 108: redundant

PR #781, opened 2021-03-28
Improve expression generator documentation

dantleech:
- Add a sphinx extension to load the "approval" examples
- The example config, command and output are tested as approval tests


dantleech:
@OskarStark hey - i've added a new custom directive `approved` - which is failing with the same issue as `codeimport` did originally. the temporary fix last time was to ignore `codeimport` upstream.

Maybe we can make `BlankLineAfterDirective.php` configurable and pass in a list of directives to ignore?

dantleech:
I disabled that specific rule for now

PR #783, opened 2021-03-29
Better rendering of `rstdev`

dantleech:
The relative deviation will now be rendered specifically and with a gradient (white good => red bad).


lib/Expression/Ast/PercentageNode.php
```diff
@@ -14,8 +16,22 @@ public function __construct(Node $value)
         $this->value = $value;
     }
 
-    public function value(): Node
+    public function valueNode(): Node
     {
         return $this->value;
     }
+
+    public function value(): float
```
dantleech, line 24: the method on this interface `NumberNode` should probably be `numericValue`

PR #808, opened 2021-04-13
Renaming

dantleech:
This PR aims to stabalize the configuration options.

- All options prefixed with the extension to which they belong eg. `report.generators` instead of `reports`, `runner.bootstrap` instead of `bootstrap`


dantleech:
Note the next release will be BETA to avoid breaks with `@alpha`

PR #813, opened 2021-04-20
Automatically creates missing dirs and refactors payload test

alexandrajulius:
I wanted to write  a regression test for the change, but it's a private method and the created file is temporary anyway. 
So I started refactoring your tests in order to iteratively understand your code.
**prophesize()** is deprecated and will be removed in PHPUnit 10 https://phpunit.de/supported-versions.html
I mocked `ProcessFactory` now, actually Sebastian Bergmann only wants you to mock interfaces. I didn't create an Interface for `ProcessFactory` now. Maybe you don't want to mock `ProcessFactory` but instantiate it, because you want to test its behaviour too? Otherwise, we could think about creating an Interface for it. But just a suggestion.

### DIFF
--- a/tests/Unit/Remote/PayloadTest.php
+++ b/tests/Unit/Remote/PayloadTest.php
@@ -15,7 +15,8 @@ namespace PhpBench\Tests\Unit\Remote;
 use PhpBench\Remote\Payload;
 use PhpBench\Remote\ProcessFactory;
 use PhpBench\Tests\TestCase;
-use Prophecy\Argument;
+use PHPUnit\Framework\MockObject\Invocation;
+use PHPUnit\Framework\MockObject\Rule\InvocationOrder;
 use RuntimeException;
 use Symfony\Component\Process\Process;
 
@@ -26,8 +27,8 @@ class PayloadTest extends TestCase
 
     protected function setUp(): void
     {
-        $this->process = $this->prophesize(Process::class);
-        $this->processFactory = $this->prophesize(ProcessFactory::class);
+        $this->process = $this->createMock(Process::class);
+        $this->processFactory = $this->createMock(ProcessFactory::class);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


tests/Unit/Remote/PayloadTest.php
```diff
@@ -26,8 +26,8 @@ class PayloadTest extends TestCase
 
     protected function setUp(): void
     {
-        $this->process = $this->prophesize(Process::class);
-        $this->processFactory = $this->prophesize(ProcessFactory::class);
+        $this->process = $this->createMock(Process::class);
+        $this->processFactory = $this->createMock(ProcessFactory::class);
     }
```
dantleech, line 31: instead of this you can:

```
use ProphecyTrait;
```

this is the approach going forward

alexandrajulius: Ah, you are using phpspec and phpunit together. I was aiming to keep maintainability effort low by simplifying the test to use phpunit only.

dantleech: I think in general I'd like to use the native PHPUnit mock framework, but would rather do this all together rather than mixing the two approaches. Using prophecy and that trait means nothing needs to change. The PHPUnit approach doesn't look as bad as I remember though ,,,

alexandrajulius: I'd still say, you don't want to mock `ProcessFactory` but instantiate it.

alexandrajulius: > but would rather do this all together rather than mixing the two approaches
You are mixing two approaches currently (and one of them is deprecated).
I think it's more realistic to improve software over time, not at once.
It gives you the freedom to apply changes iteratively, just as you need them.

dantleech: > I'd still say, you don't want to mock ProcessFactory but instantiate it.

We need to mock the ProcessFactory (this is why it exists) there could indeed be an interface.

> You are mixing two approaches currently

No, we only use prophecy so far (AFAICS)

alexandrajulius:
Closing this PR since it contains unwanted changes.

Issue #818, opened 2021-04-21
 DOMElement::__construct(): unterminated entity reference


staabm:
running a benchmark, which emits a `&` characters lead to the following error

```php
class MyBench {
    /**
     * @return void
     *
     * @Revs(5)
     * @Iterations(3)
     */
    public function benchAmpersand() {
        echo 'Fix & Foxy';
    }
}
```

```
In Element.php line 29:

  [ErrorException]
  Warning: DOMElement::__construct(): unterminated entity reference  FoxyFix & FoxyFix & FoxyFix & FoxyFix & Foxy


Exception trace:
  at C:\dvl\GitHub\philipp\vendor\phpbench\dom\lib\Element.php:29
 PhpBench\Dom\Element->appendElement() at C:\dvl\GitHub\philipp\vendor\phpbench\phpbench\lib\Serializer\XmlEncoder.php:131
 PhpBench\Serializer\XmlEncoder->processVariant() at C:\dvl\GitHub\philipp\vendor\phpbench\phpbench\lib\Serializer\XmlEncoder.php:102
 PhpBench\Serializer\XmlEncoder->processSubject() at C:\dvl\GitHub\philipp\vendor\phpbench\phpbench\lib\Serializer\XmlEncoder.php:85
 PhpBench\Serializer\XmlEncoder->processBenchmark() at C:\dvl\GitHub\philipp\vendor\phpbench\phpbench\lib\Serializer\XmlEncoder.php:72
 PhpBench\Serializer\XmlEncoder->encode() at C:\dvl\GitHub\philipp\vendor\phpbench\phpbench\lib\Console\Command\Handler\DumpHandler.php:43
 PhpBench\Console\Command\Handler\DumpHandler->dumpFromInput() at C:\dvl\GitHub\philipp\vendor\phpbench\phpbench\lib\Console\Command\RunCommand.php:115
 PhpBench\Console\Command\RunCommand->execute() at C:\dvl\GitHub\philipp\vendor\symfony\console\Command\Command.php:256
 Symfony\Component\Console\Command\Command->run() at C:\dvl\GitHub\philipp\vendor\symfony\console\Application.php:971
 Symfony\Component\Console\Application->doRunCommand() at C:\dvl\GitHub\philipp\vendor\symfony\console\Application.php:290
 Symfony\Component\Console\Application->doRun() at C:\dvl\GitHub\philipp\vendor\symfony\console\Application.php:166
[...]

PR #819, opened 2021-04-21
added failing XmlEncoder testcase regarding ampersand

staabm:
closes https://github.com/phpbench/phpbench/issues/818

requires https://github.com/phpbench/dom/pull/3

### DIFF
--- a/lib/Serializer/XmlEncoder.php
+++ b/lib/Serializer/XmlEncoder.php
@@ -127,7 +127,7 @@ class XmlEncoder
             $errorsEl = $variantEl->appendElement('errors');
 
             foreach ($variant->getErrorStack() as $error) {
-                $errorEl = $errorsEl->appendElement('error', $error->getMessage());
+                $errorEl = $errorsEl->appendTextNode('error', $error->getMessage());
                 $errorEl->setAttribute('exception-class', $error->getClass());
                 $errorEl->setAttribute('code', $error->getCode());
                 $errorEl->setAttribute('file', $error->getFile());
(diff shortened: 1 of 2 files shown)


staabm:
hmm it seems this now reproduces my problem, but I guess the fix should actually be in https://github.com/phpbench/dom ?

dantleech:
> hmm it seems this now reproduces my problem, but I guess the fix should actually be

yes - looks like there needs to be another API method `createTextElement` ([see here](https://www.php.net/manual/en/domdocument.createelement.php#refsect1-domdocument.createelement-notes)).

Alternatively could refactor to use `$errorsEl->appendChild($dom->createTextElement('foo', 'bar'))` instead, but would be more consistent to add a new method to `phpbench/dom`

dantleech:
Tagged and bumped the dom lib, so rebase should make this green.

lib/Serializer/XmlEncoder.php
```diff
@@ -61,10 +61,9 @@ public function encode(SuiteCollection $suiteCollection): Document
                 $infoEl = $envEl->appendElement($information->getName());
 
                 foreach ($information as $key => $value) {
-                    $valueEl = $infoEl->appendElement('value');
+                    $valueEl = $infoEl->appendTextNode('value', $value);
```
dantleech, line 64: note this is just one place - in your original example I think the `echo "Foo & Bar"` was in the error message. As none of what we do here should be passing a literal `&` is is probably fine to replace all `appendElement` with `appendTextNode`

staabm: hmm I did not spot any other occurence where a node is appended with a value.

all other remaining spots seem to only set attributes. did I miss something?

staabm: ahh just saw 2 other spots. anything left?

dantleech:
Great, thanks for this!

PR #841, opened 2021-05-19
HTML Report 2

dantleech:
This PR replaces #839 .

It uses an _object renderer_ to render the report objects directly with PHP templates.

- All templates are named after the classes they represent.
- One template has one object
- The renderer accepts an array of template paths enabling all or specific templates to be overridden.

TODO:

- [x] Introduce an HtmlView providing links to the CSS etc.
- [x] The printers contain quite complex logic to f.e. convert the absolute time value to a given unit. This could be reconciled with the templating system (e.g. using the templating engine as a node printer)


lib/Expression/Ast/VariableNode.php
```diff
@@ -2,7 +2,7 @@
 
 namespace PhpBench\Expression\Ast;
 
-class VariableNode implements Node
+final class VariableNode extends Node
```
dantleech, line 5: use inheritence with the AST nodes - this helps with template resolution, but also would have been a better idea from the start.

lib/Extension/TemplateExtension.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace PhpBench\Extension;
+
+use PhpBench\DependencyInjection\Container;
+use PhpBench\DependencyInjection\ExtensionInterface;
+use PhpBench\Template\ObjectPathResolver\ReflectionObjectPathResolver;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class TemplateExtension implements ExtensionInterface
```
dantleech, line 10: remove

lib/Report/Renderer/TemplateRenderer.php
```diff
@@ -0,0 +1,61 @@
...
+            mkdir($dirname, 0777, true);
+        }
+
+        if (false === file_put_contents($outPath, $rendered)) {
+            throw new RuntimeException(sprintf(
+                'Could not write report to file "%s"', $outPath
+            ));
+        }
+        $this->output->writeln(sprintf('Written report to: %s', $outPath));
+    }
```
dantleech, line 60: add tests for dir creation etc.

lib/Template/ObjectPathResolver/ReflectionObjectPathResolver.php
```diff
@@ -0,0 +1,70 @@
...
+
+    private function classToPath(string $classFqn): string
+    {
+        foreach ($this->prefixMap as $prefix => $pathPrefix) {
+            if (false !== strpos($classFqn, $prefix)) {
+                return sprintf('%s/%s.phtml', rtrim($pathPrefix, '/'), ltrim(str_replace('\\', '/', substr($classFqn, strlen($prefix))), '/'));
+            }
+        }
+
+        throw new CouldNotResolvePath(sprintf(
```
dantleech, line 64: this exception is an artficat from an earlier iteration, remove it.

lib/Report/Renderer/TemplateRenderer.php
```diff
@@ -0,0 +1,61 @@
...
+    {
+        $rendered = $this->renderer->render($report);
+        $dirname = dirname($this->outputDir);
+        $outPath = $this->outputDir . '/index.html';
+
+        if (!file_exists($dirname)) {
+            mkdir($dirname, 0777, true);
+        }
+
+        if (false === file_put_contents($outPath, $rendered)) {
```
dantleech, line 54: Write to STDOUT by default? depends if this renderer would be responsible for rendering mutliple pages or not - STDOUT output could be done later in anycase.

docs/configuration.rst
```diff
@@ -438,6 +438,39 @@ Default: ``[]``
...
+Default: ``["\/home\/daniel\/www\/phpbench\/phpbench\/lib\/Extension\/..\/..\/templates"]``
+
+Types: ``["array"]``
+
+.. _configuration_report_html_output_dir:
+
+report.html_output_dir
+~~~~~~~~~~~~~~~~~~~~~~
+
+Path in which to render HTML reports
```
dantleech, line 457: This is either a global default, or a per-report setting - or both. This could be overridden at the report level whilst providing a default output dir.

lib/Template/Expression/Printer/TemplatePrinter.php
```diff
@@ -0,0 +1,65 @@
...
+        $this->nodePrinters = $nodePrinters;
+        $this->renderer = $renderer;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function print(Printer $printer, Node $node): string
+    {
+        $hash = serialize($node);
```
dantleech, line 45: Is there a better way to cache by value?

lib/Template/TemplateService/MappedTemplateService.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace PhpBench\Template\TemplateService;
+
+use PhpBench\Template\TemplateService;
+use RuntimeException;
+
+class MappedTemplateService implements TemplateService
```
dantleech, line 8: final

PR #845, opened 2021-05-28
Refactor parameters

dantleech:
Parameter handling is a bit of a mess currently, and there is an issue introduced in 1.0 where foreign PHP objects can be introduced into PHPBenchs runtime. This refactoring aims to facilitate the introduction of a ParameterEnvelope to isolate parameters from the PHPBench process while also retaining the ability to see and query scalar values in reports.


examples/Command/report-generators-break
```diff
@@ -14,6 +14,6 @@ NothingBench
 +--------------+-----+------+-----+----------+----------+--------+
 | subject      | set | revs | its | mem_peak | mode     | rstdev |
 +--------------+-----+------+-----+----------+----------+--------+
-| benchNothing | 0   | 1    | 1   | 100b     | 10.000μs | ±0.00% |
+| benchNothing |     | 1    | 1   | 100b     | 10.000μs | ±0.00% |
```
dantleech, line 17: side effect which makes more sense really.

lib/Model/ParameterSet.php
```diff
@@ -12,25 +12,30 @@
...
+    private $parameters;
+
     /**
      * @param array<string,mixed> $parameters
      */
     public function __construct(string $name, array $parameters = [])
     {
         $this->name = $name;
-        parent::__construct($parameters);
+        $this->parameters = $parameters;
```
dantleech, line 38: Add validation here

lib/Executor/Benchmark/template/remote.template
```diff
@@ -12,6 +12,9 @@ $beforeMethods = {{ beforeMethods }};
 $afterMethods = {{ afterMethods }};
 $bootstrap = '{{ bootstrap }}';
 $warmup = {{ warmup }};
+$parameters = array_map(function (string $value) {
+    return unserialize($value);
```
dantleech, line 16: could make this b/c for anybody that created a custom template?

dantleech: need to make "safe" parameter opt-in and use a new template

extensions/xdebug/tests/Unit/XDebugUtilTest.php
```diff
@@ -35,7 +36,7 @@ public function testGenerate($class, $subject, $expected): void
         $params = [
             'classPath' => '/foobar',
             'parameterSetName' => '7',
-            'parameters' => ['asd'],
+            'parameters' => ParameterSet::fromUnwrappedParameters('test', ['asd']),
```
dantleech, line 39: possible b/c break

lib/Benchmark/RunnerConfig.php
```diff
@@ -165,9 +165,11 @@ public function getWarmup(array $default = null): ?array
...
      * Override parameters.
      *
+     * @param parameters $default
+     *
      * @return mixed[]
      */
-    public function getParameterSets($default = null)
+    public function getParameterSets(?array $default = null)
     {
         $parameters = $this->parameters ? [[$this->parameters]] : $default;
```
dantleech, line 174: property is not type hinted for `null`

lib/Executor/Benchmark/template/remote.template
```diff
@@ -25,21 +28,21 @@ $benchmark = new $class();
...
 foreach ($beforeMethods as $beforeMethod) {
-    $benchmark->$beforeMethod({{ parameters }});
+    $benchmark->$beforeMethod($parameters);
 }
 
 // run warmup if required
 if ($warmup) {
     for ($i = 0; $i < $warmup; $i++) {
-        $benchmark->{{ subject }}({{ parameters }});
+        $benchmark->{{ subject }}();
```
dantleech, line 37: missing params

lib/Executor/Benchmark/template/remote.template
```diff
@@ -25,21 +28,21 @@ $benchmark = new $class();
...
     }
 }
 
 // run benchmark
 $startTime = microtime(true);
 
 for ($i = 0; $i < {{ revolutions }}; $i++) {
-    $benchmark->{{ subject }}({{ parameters }});
+    $benchmark->{{ subject }}($parameters);
```
dantleech, line 45: passing $parameters probably incurs an overhead, have two loops one with and one without

lib/Model/ParameterSet.php
```diff
@@ -12,33 +12,25 @@
 
 namespace PhpBench\Model;
 
-use ArrayObject;
-
-/**
- * @extends ArrayObject<string,mixed>
- */
-class ParameterSet extends ArrayObject
+final class ParameterSet
```
dantleech, line 15: revert final, make it `/** final */`

lib/Model/ParameterSet.php
```diff
@@ -12,33 +12,25 @@
...
-        parent::__construct($parameters);
-    }
+    private $parameters;
 
     /**
-     * @param array<string,mixed> $parameters
+     * @param array<string,ParameterContainer> $parameters
      */
-    public static function create(string $name, array $parameters): self
+    private function __construct(string $name, array $parameters)
```
dantleech, line 30: make it public for b/c add comment

lib/Model/ParameterSet.php
dantleech: need to "break" B/C on this class anyway as the constructor requires different arguments.

PR #850, opened 2021-05-29
POC Overview Report

dantleech:
This currently non-configurable report is a POC for chart rendering. It generates an overview table, and then a overview chart, then a chart for each benchamark, comparing however many `--ref` you provide

image

It would be good to be able to add the existing table reports along side the charts, and to this end the logic to generate that table from the DataFrame needs to be extracted from the ExpressionGenerator.

Things to do to introduce this:

- Create an additional class of report "components" which accept a data frame: these configurable components can then be used by the existing ReportGenerators.
- Apply the refactorings suggested in this PR.
  - Introduce the ReportBuilder and the change to make the Report accept any Report Component (i.e. not just tables)
  - Introduce a facade for the EL which can evaluate an given expression without first parsing it.
  - The consle renderer currently knows only how to render Tables, refactor to permit rendering different components (console bar charts can come later)
  -


dantleech:
Replaced by #853

PR #851, opened 2021-05-29
Component Ggenerator

dantleech:
This PR introduces a new report generator: the `Component` generator.

`Components` are small elements of a report (e.g. a Table, BarChart) etc. and enable reports to be composable.

Components Generators are similar to report Generators, but they accept a `DataFrame` instead of the `SuiteCollection` and return a `Component` (e.g. a Table, BarChart, etc). 

This will enable reports such as #850 to be produced through configuration and permit reuse of the type of table available in the current `expression` generator, but the configuration is different (the column names are not presented as key => series, but instead exposed only through the data frame).

This method of report generation will probably replace the `expression` generator in 2.0.

```
        "component": {
            "generator": "component",
            "partition": ["benchmark_class"],
            "title": "first(frame[\"benchmark_name\"])",
            "components": [
                {
                    "_type": "table_aggregate",
                    "title": "first(frame[\"benchmark_name\"])",
                    "partition": ["subject_name", "variant_name"],
                    "row": {
                        "subject": "first(partition[\"subject_name\"])",
                        "mode": "mode(partition[\"result_time_avg\"]) as ms"
                    }
                },
                {
                    "_type": "table_aggregate",
                    "partition": ["iteration_index"],
                    "row": {
                        "subject": "first(partition[\"subject_name\"])",
                        "mode": "mode(partition[\"result_time_avg\"]) as ms"
                    }
                }
            ]
        }
  ```

TODO:

- [x] Make the ComponentSuiteGeneator implements the ComposerGeneratorInterface so that the configuration can be shifted up one level.
- [x] Rename `table` component to `aggregate_table` (as it operates by aggregating partitions) and a `table` generator could have statis columns.
- [x] Find a good way to test
- [x] Generally review the naming etc

### CODE REVIEW

File: lib/Extension/ExpressionExtension.php [hunk 1]
```diff
@@ -337,6 +338,12 @@ public function load(Container $container): void
                 new RealExpressionLanguage($container->get(Lexer::class), $container->get(Parser::class))
             );
         });
+        $container->register(ExpressionEvaluator::class, function (Container $container) {
```
dantleech, line 341: This could be used to simplify other classes which currently require both the EL and the evaluator.

lib/Extension/ReportExtension.php
```diff
@@ -53,6 +58,7 @@ class ReportExtension implements ExtensionInterface
     public const PARAM_TEMPLATE_MAP = 'report.template_map';
     public const PARAM_TEMPLATE_PATHS = 'report.template_paths';
     public const PARAM_OUTPUT_DIR_HTML = 'report.html_output_dir';
+    const TAG_COMPONENT_GENERATOR = 'report.component_generator';
```
dantleech, line 61: ...

lib/Report/Model/Report.php
```diff
@@ -15,31 +17,36 @@ final class Report
...
      */
-    private $tables;
+    private $objects;
 
     /**
-     * @param Table[] $tables
+     * @param ComponentInterface[] $objects
      */
-    private function __construct(array $tables, ?string $title, ?string $description)
+    public function __construct(array $objects, ?string $title, ?string $description)
```
dantleech, line 27: ADd comment here, should not be instantiated directly

PR #853, opened 2021-05-30
Barchart Component

dantleech:
Introduces an opinonated BarChart component:
 
image

Further work that is needed:

- [x] Ability to use the expression language in labels? e.g. `label: "Benchmark: %%partition["benchmark_name"]%%` or make labels expressions only `label: "\"Benchmark: \" ~ partition[\"benchmark_name\"]"`
- [x] How to usefully display multiple suites in a table? 
    - Dynamically add columns (how?)
    - Assign colors to the partitioned sets?
    - :heavy_check_mark: tabs - later
- [ ] Console renderer will not render `component` reports at all: needs to be refactored to enable it to render things other than tables.
- [x] Depending on a CDN for Plotly and Bootstrap is probably fine to start with, but would be good to bundle them or enable them to be hosted locally.


templates/html/HtmlDocument.phtml
```diff
@@ -4,6 +4,7 @@
         <meta name="viewport" content="width=device-width, initial-scale=1.0" />
         <title><?php echo $object->title() ?></title>
         <link href="https://cdn.jsdelivr.net/npm/bootstrap@4.6.0/dist/css/bootstrap.min.css" rel="stylesheet">
+        <script src="https://cdn.plot.ly/plotly-latest.min.js"></script>
```
dantleech, line 7: probably want to lock this

lib/Expression/ExpressionEvaluator.php
```diff
@@ -29,4 +31,21 @@ public function evaluate(string $expression, array $params): Node
...
+     *
+     * @param parameters $params
+     */
+    public function evaluatePhpValue(string $expression, array $params)
+    {
+        $node = $this->evaluator->evaluateType($this->language->parse($expression), PhpValue::class, $params);
+
+        // TODO: ListNode does not return the PHP value
+        if ($node instanceof ListNode) {
+            return $node->phpValues();
```
dantleech, line 46: Can probably fix this without breaking B/C

lib/Report/ComponentGenerator/BarChartComponentGenerator.php
```diff
@@ -0,0 +1,72 @@
...
+     * {@inheritDoc}
+     */
+    public function configure(OptionsResolver $options): void
+    {
+        $options->setDefaults([
+            self::PARAM_X_PARTITION => [],
+            self::PARAM_SET_PARTITION => [],
+        ]);
+        $options->setRequired(self::PARAM_Y_EXPR);
+        $options->setRequired(self::PARAM_Y_ERROR_MARGIN);
```
dantleech, line 40: Could make error margins optional.

lib/Report/Model/BarChartDataSet.php
```diff
@@ -0,0 +1,36 @@
...
+     */
+    public $xSeries;
+    /**
+     * @var mixed[]
+     */
+    public $ySeries;
+    /**
+     * @var mixed[]
+     */
+    public $errorMargins;
```
dantleech, line 22: If going the pulic property way then change the Table components to do this too (although possibly there is a use case for having richer models when rendering on the console f.e.)

PR #854, opened 2021-05-30
Fix duplicated key "runner.php_config" on phpbench.json.dist

Tito-Kati:
The `phpbench.json.dist` file has the `"runner.php_config"` key duplicated.

Technically the JSON specification does not make any mention of duplicate keys being invalid or valid, but this could cause problems in some JSON libraries implementations.


Chemaclass (approved):

dantleech:
Thanks!

PR #856, opened 2021-06-02
Console renderer

dantleech:
Refactor the console renderer to allow rendering of individual components.


lib/Report/Renderer/ConsoleRenderer.php
```diff
@@ -47,55 +37,14 @@ public function __construct(OutputInterface $output, Printer $printer)
...
-    }
-
-    private function buildRows(PhpBenchTable $table): array
-    {
-        return array_map(function (TableRow $row) {
-            return array_map(function (Node $node) {
-                return $this->printer->print($node);
-            }, $row->cells());
-        }, $table->rows());
+        $options->setDefault('table_style', null);
```
dantleech, line 48: This option is not propagated to the indivdual renderers. It should be marked as dead and the responsiblity moved to the table generator (??).

PR #857, opened 2021-06-02
Update PHP CS Fixer and PSR12

Tito-Kati:
- Updated PHP CS Fixer to its latests version (v3.0.0).
- [PSR2](https://www.php-fig.org/psr/psr-2/) is deprecated since 2019-08-10 so I updated the project to PSR12.


MGatner (approved):

dantleech:
Thanks for this, cherry-picked and applied the CS again here #915

PR #861, opened 2021-06-06
Tabbed components

dantleech:
Add a `tabbed` option to the `report` component, to render each component in a tab (where supported by the renderer)

image


lib/Report/Generator/ComponentGenerator.php
```diff
@@ -117,6 +121,10 @@ public function generateComponent(DataFrame $dataFrame, array $config): Componen
             }
         }
 
+        if ($config[self::PARAM_TABBED]) {
+            $builder = $builder->enableTabs();
```
dantleech, line 125: builder is mutable, remove assignment

PR #862, opened 2021-06-06
Fix mode() when population has NULL values

dantleech:
Need to consider this and apply it also to other aggregate functions.

### DIFF
--- a/lib/Expression/Func/ModeFunction.php
+++ b/lib/Expression/Func/ModeFunction.php
@@ -11,6 +11,6 @@ final class ModeFunction
 {
     public function __invoke(ListNode $values, ?IntegerNode $space = null): FloatNode
     {
-        return new FloatNode(Statistics::kdeMode($values->phpValues(), $space ? $space->value() : 512));
+        return new FloatNode(Statistics::kdeMode($values->nonNullPhpValues(), $space ? $space->value() : 512));
     }
 }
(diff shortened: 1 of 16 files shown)


lib/Report/Generator/ComponentGenerator.php
```diff
@@ -80,7 +80,7 @@ public function configure(OptionsResolver $options): void
      */
     public function generate(SuiteCollection $collection, Config $config): Reports
     {
-        $dataFrame = $this->transformer->suiteToFrame($collection);
+        $dataFrame = $this->transformer->suiteToFrame($collection, true);
```
dantleech, line 83: This isn't related - it's required to show referenced benchmarks when `run` is used instead of `report` (investigate why it  worked with this as `false`)

lib/Expression/Func/ModeFunction.php
```diff
@@ -11,6 +11,6 @@ final class ModeFunction
 {
     public function __invoke(ListNode $values, ?IntegerNode $space = null): FloatNode
     {
-        return new FloatNode(Statistics::kdeMode($values->phpValues(), $space ? $space->value() : 512));
+        return new FloatNode(Statistics::kdeMode($values->nonNullPhpValues(), $space ? $space->value() : 512));
```
dantleech, line 14: Might be better to extract this to a separate function in `Math`

Issue #643, opened 2020-06-25
How to ignore certain annotations?


shmax:
Hi there, our code base uses some custom annotations that are specific to the phpstan library, for example:

```php
/**
 * @template T of Node
 * @phpstan-implements ArrayAccess<int|string, T>
 * @phpstan-implements IteratorAggregate<T>
 */
class NodeList implements ArrayAccess, IteratorAggregate, Countable
{

```

These seem to be giving phpbench fits. It complains with
```
 [Semantical Error] The annotation "@template" in benchmark: \GraphQL\Language\AST\NodeList was never imported. Did you maybe forget to add a "use" statement for this annotation? 
```

I've been poring over documentation and diving into code for an hour, and for the life of me I can't figure out how to configure phpbench with knowledge of these and other custom annotations.

And advice greatly appreciated!

dantleech:
You can tell PHPBench not to use global imports with `"annotation_import_use" => false` in `phpbench.json`. You can then import the full FQNs of the PHPBench annotations, e.g. `use PhpBench\Benchmark\Metadata\Annotations\Sleep`

But I'd question why PHPBench is loading a`NodeList` class?

Typically I'd expect PHPBench files to live in a tests folder, e.g. `tests\Bench` and by default it only loads `*Bench.php` files?

dantleech:
Ah, and I guess `annotation_import_use` won't help in case of these tags which, if anything, should be accounted for [here](https://github.com/phpbench/phpbench/blob/master/lib/Benchmark/Metadata/AnnotationReader.php) (maybe it could be made to be configurable).

shmax:
Thanks much for the reply!

> "annotation_import_use" => false in phpbench.json

Is this really the _de facto_ way to do this? A search on github for that key only turns up 5 matches, and none of them are in json files.

> You can then import the full FQNs of the PHPBench annotations, e.g. use PhpBench\Benchmark\Metadata\Annotations\Sleep

That doesn't have meaning in the case of `@template`. There is no matching class by that name; it's just a custom annotation that the phpstan parser knows to watch for.

> But I'd question why PHPBench is loading aNodeList class?
I don't know, it's just one of the classes that is exercised in the code being benchmarked. Is phpbench not supposed to be parsing the phpdoc blocks of the files in the code it is testing?

> tags which, if anything, should be accounted for here
[...]

PR #883, opened 2021-07-01
Fix benchmark finder to only pick up *Bench files

dantleech:
This "bug" has been present since 2016.

It is [documented](https://phpbench.readthedocs.io/en/latest/quick-start.html#create-a-benchmark) that we only run files suffixed with `Bench.php` and this was completely my understanding, however since version 0.10.0 PHPBench will take into account any PHP file in the benchmark `path`.

This is not good because:

- It will try and read annotations from non-benchmark classes (https://github.com/phpbench/phpbench/issues/643)
- People who read the documentation (or listened to anything I said) would not expect this behavior.
- The "reflector" will execute code in non `*Bench` files, which increases the chance of unintentional code execution.

Fixing it however means a non-fatal change in behavior:

- Anybody depending on benchmark classes not having the suffix will need to set the new option `runner.file_pattern` option to `*.php`

---

This PR adds the option `runner.file_pattern` but sets it to NULL by default to preserve the existing behavior. If it is NULL and a benchmark file does not end in `Bench.php` a warning is displayed:

image

### DIFF
--- a/lib/Benchmark/BenchmarkFinder.php
+++ b/lib/Benchmark/BenchmarkFinder.php
@@ -128,13 +128,14 @@ class BenchmarkFinder
         }
 
         foreach ($finder as $file) {
-            if ($this->benchPattern === null && substr($file->getPathInfo(), 0, -9) !== 'Bench.php') {
+            assert($file instanceof SplFileInfo);
+
+            if ($this->benchPattern === null && substr($file->getFilename(), -9) !== 'Bench.php') {
                 $this->logger->warning(sprintf(
                     'File "%s" has been identified as a benchmark file but it does not end with ' .
-                    '`Bench.php` and no `runner.file_pattern` has been specified. ' .
-                    'It is recommended to set a file pattern. In PHPBench 2.0 the pattern will be ' .
-                    '`*Bench.php` by default.',
-                    $file->getPathInfo()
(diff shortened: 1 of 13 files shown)


dantleech:
cc @greg0ire

dantleech:
Maybe better to leave the default file pattern as `*.php` in a bug fix release, and add `*Bench.php` in 1.1

lib/Benchmark/BenchmarkFinder.php
```diff
@@ -35,10 +35,16 @@ class BenchmarkFinder
...
+    /**
+     * @var string
+     */
+    private $benchPattern;
+
+    public function __construct(MetadataFactory $factory, string $cwd, string $benchPattern = null)
     {
         $this->factory = $factory;
         $this->cwd = $cwd;
+        $this->benchPattern = $benchPattern ?: '*.php';
```
dantleech, line 47: remove default

greg0ire:
Not sure if running the scripts should be considered a security issue, meaning I don't see how a malicious actor could use that to their advantage. They would contribute a malicious script and a dependency to phpbench? Sounds far fetched. But there is a potential for disaster when running random scripts, so there is a big bug.

dantleech:
tagged phpbench 1.0.3

PR #889, opened 2021-07-14
Allow an expression to be specified for evaluating the X-Label in Bar Charts

dantleech:


lib/Report/Model/BarChart.php
```diff
@@ -118,4 +130,12 @@ public function yAxisLabelRange(int $steps): array
 
         return range(0, $max, $step);
     }
+
+    /**
+     * @return scalar[]
+     */
+    public function xLabels(): array
+    {
+        return array_replace($this->xAxes(), $this->xLabels ?: []);
```
dantleech, line 139: test

lib/Report/ComponentGenerator/BarChartAggregateComponentGenerator.php
```diff
@@ -67,53 +71,62 @@ public function configure(OptionsResolver $options): void
     public function generateComponent(DataFrame $dataFrame, array $config): ComponentInterface
     {
         $xSeries = [];
+        $xAxes = [];
         $xLabels = [];
         $errorMargins = [];
         $ySeries = [];
+        $xOffset = -1;
```
dantleech, line 78: remove

PR #894, opened 2021-07-21
Console Barchart labels

dantleech:
Show legend for X ticks

image


lib/Report/ComponentGenerator/BarChartAggregateComponentGenerator.php
```diff
@@ -180,7 +180,9 @@ private function partitionFunction(array $partitionColumns): Closure
                 $hash[] = $this->evaluator->evaluatePhpValue($column, $row->toRecord());
             }
 
-            return implode('-', $hash);
+            return implode('-', array_filter($hash, function (string $value) {
+                return $value !== '';
+            }));
```
dantleech, line 185: prevent trailing dash if segment is ''

PR #898, opened 2021-07-25
Fix performance issue

dantleech:
This PR fixes the performance bottleneck which was caused by always converting primitive values to nodes and back again during evaluation. Now, if a ListNode is instantiated from primitives, it does not eagerly convert them to nodes.

This results in a x4 speed improvement when compared to `1.0.x`

### DIFF
--- a/lib/Expression/Ast/ListNode.php
+++ b/lib/Expression/Ast/ListNode.php
@@ -4,20 +4,53 @@ namespace PhpBench\Expression\Ast;
 
 final class ListNode extends DelimitedListNode
 {
+    private $values;
+
     /**
      * @param array<mixed> $values
      */
     public static function fromValues(array $values): self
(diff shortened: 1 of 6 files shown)


lib/Expression/Ast/ListNode.php
```diff
@@ -2,22 +2,72 @@
...
+/**
+ * Optimisation note: When `fromValues` is used, the source of truth is
+ * switched from a list of nodes to a list of primitive values. This avoids the
+ * overhead of converting nodes back to values and improves performance
+ * significantly.
+ */
 final class ListNode extends DelimitedListNode
 {
+    /**
+     * @var mixed[]
```
dantleech, line 14: |null

dantleech:
Didn't fix the issue

PR #899, opened 2021-07-25
Fix links to documentation.

drupol:


dantleech:
Hey, is there a reason? the `.org` works fine?

drupol:
> Hey, is there a reason? the `.org` works fine?

Indeed, I just noticed that now it works. Most probably that it was not working yesterday evening when I did the PR.

Anyway, this PR is not totally invalid because the `.org` redirected to the `.io`, so I guess it's better to align the README to using `.io` everywhere instead of having a mix of `.io` and `.org`.

dantleech:
Thanks

PR #901, opened 2021-07-29
Update and rename AttriutedBench.php to AttributedBench.php

rustamwin:


dantleech (approved):

dantleech:
Thanks, this test is referenced in the "Annotributes" documentation I think and that would also need updating.

rustamwin:
> Thanks, this test is referenced in the "Annotributes" documentation I think and that would also need updating.

Done.

dantleech:
Thanks

PR #904, opened 2021-07-31
POC: Opcode counter

dantleech:
This PR experiments with adding an `opcode` executor which provides a count of the opcodes executed during the benchmark.

**TL;DR**: It should be possible to provide a count of the opcodes for the subject only.

## Counting Opcodes

- It seems not possible to enable the opcode debugging during execution (e.g. to capture only the subject's opcodes)
- It *would be possible* to first execute the script without running the subject, then again with the subject, and subtract the first from the second.

## Including Opcodes alongside the regular results

Would be possible to add a new `CompositeExecutor` which ran a set of executors. I.e. to execute the opcode executor and the regular executor.

## Pre and post optimisation

By using the `CompositeExecutor` it would be possible to run two instances of the Opcode executor, one for unoptimised and one with optimised results.

## Debugging Opcodes

It would also be good to add a convenience command to debug the opcodes. However, I cannot see a way to separate the signal from the noise. The opcodes are not presented in the order that would make this possible.


lib/Expression/Exception/VariableNotFound.php
```diff
@@ -2,6 +2,6 @@
 
 namespace PhpBench\Expression\Exception;
 
-class KeyDoesNotExist extends EvaluationError
+class VariableNotFound extends EvaluationError
```
dantleech, line 5: previously unused class, technically a B/C break but :shrug:

Issue #905, opened 2021-08-03
PHP 8.1 - Namespace is not calculated correctly


agustingomes:
Hi,

I am running some preliminary tests with the beta1 version of PHP 8.1, and I found a potential problem.

The current master state of the [RemoteReflector](https://github.com/phpbench/phpbench/blob/c978079d13dd51dc75099b0451f5e7be824b24d4/lib/Reflection/RemoteReflector.php#L129) is not being able to correctly calculating the namespace, and therefore the resulting template for the benchmark run ends in error.

In 8.0 the  $tokenId matches the hardcoded `314`:
![image](https://user-images.githubusercontent.com/15325886/128065742-8440791b-4929-47d7-83f9-510b14eb5639.png)

I feel in 8.1 the token's value changed, hence resulting in that part of the namespace not being extracted correctly:
![image](https://user-images.githubusercontent.com/15325886/128065513-08b02cc7-716a-42a5-b814-644b9303fa3a.png)

dantleech:
yes can be :( this code also broke with 8.0

dantleech:
PR where this was introduced: https://github.com/phpbench/phpbench/pull/665/commits/1c2e17d30d8dc7ba20056d83e0f821c8056043aa#diff-6b0463b0983e774b73f65bfc37e8d70567eb9a7939e81e61976541661d8200e1R138

agustingomes:
@dantleech I understand now why that was the case, TL;DR 8.0 added a specific token for the qualified namespaces:

https://wiki.php.net/rfc/namespaced_names_as_token

I will open a PR with a siggestion to overcome this.

agustingomes:
Resolved with the merge of Use token for full namespace whenever available #906

PR #906, opened 2021-08-04
Use token for full namespace whenever available

agustingomes:
PHP 8.0 introduced a new token for fully qualified namespaces, represented
by the constant `T_NAME_QUALIFIED`, which in that version had the integer representation of 314.

With the upcoming version 8.1, the token received a different integer representation, thus breaking the logic with the previously hardcoded integer value.

Closes issue https://github.com/phpbench/phpbench/issues/905


dantleech:
thanks for finding this, the you need to run `./vendor/bin/php-cs-fixer` or `composer integrate`.

also wonder if it's yet possible to add 8.1 to the github actions..

agustingomes:
will apply the CS fixes later in the day.

as for the 8.1, I can take a look at it in a new branch, also later in the day

dantleech:
great, thanks for looking into this :+1:

PR #909, opened 2021-08-08
Config include

dantleech:
Allows configurations to be included by adding configuration pre processors.


docs/configuration-prelude.rst
```diff
@@ -21,3 +21,13 @@ that order and use one if it exists.
             }
         }
     }
+
+Note you can include more configuratoin files with the ``$include`` and
+``$include-glob`` directives:
```
dantleech, line 26: note that

PR #919, opened 2021-09-08
Fix/duplicate permutation runs when using abstract benchmarks

dantleech:
Final fixes for #918


tests/Benchmark/Report/Renderer/HtmlRendererBench.php
```diff
@@ -7,7 +7,6 @@
 use PhpBench\Report\Generator\OutputTestGenerator;
 use PhpBench\Report\Renderer\HtmlRenderer;
 use PhpBench\Tests\Benchmark\IntegrationBenchCase;
-use function PhpBench\Model\SuiteCollection;
```
dantleech, line 10: :thinking:

dantleech:
thanks @Ocramius :tada:  tagged 1.1.1

PR #920, opened 2021-09-11
Enable 8.1 in GH actions

dantleech:
This PR adds a CI run for PHP 8.1 and drops support for PHP 7.2


dantleech:
xdebug profile generation isn't working as expected with 8.1

PR #923, opened 2021-09-24
Fix PHP 8.1 deprecations

Crell:
I am not 100% certain of all of these, but this at least gets my tests locally to pass, except for those that require the git binary.  My container doesn't have git in it so that fails.  At least I don't get any deprecation warnings.

This works only with the updated Dom library from this PR: https://github.com/phpbench/dom/pull/4

(I have edits allowed, so feel free to fix up anything you'd like done differently. I won't take offense. :smile: )

### DIFF
--- a/lib/Remote/Payload.php
+++ b/lib/Remote/Payload.php
@@ -119,7 +119,7 @@ class Payload
 
     public function setPhpPath($phpPath): void
     {
-        $this->phpPath = escapeshellarg($phpPath);
+        $this->phpPath = $phpPath ? escapeshellarg($phpPath) : '';
     }
 
     public function disableIni(): void
--- a/lib/Color/Gradient.php
+++ b/lib/Color/Gradient.php
@@ -70,7 +70,7 @@ use function array_fill;
         }
 
         $colors = array_merge($this->colors, array_map(function (float $r, float $g, float $b) {
-            return DeprecatedColor::fromRgb(floor($r), floor($g), floor($b));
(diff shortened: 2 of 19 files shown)


lib/Remote/Payload.php
```diff
@@ -119,7 +119,7 @@ public function mergePhpConfig(array $phpConfig): void
 
     public function setPhpPath($phpPath): void
     {
-        $this->phpPath = escapeshellarg($phpPath);
+        $this->phpPath = $phpPath ? escapeshellarg($phpPath) : '';
```
dantleech, line 122: is this a phpstan issue or php 8.1 ? should we throw an exception if the string is empty?

dantleech:
Thanks, i'll look into the above comment

lib/Remote/Payload.php
Crell: I didn't want to add types to the signature as that would be an API change.  That is probably the better solution in general, but I was going for the absolute minimum impact to make the errors go away as I didn't have the broader context.

Thanks for the merge. I look forward to a release so that I can benchmark a new 8.1 library I'm working on. :smile:

dantleech: ideally I want to get GH actions running with 8.1 before tagging, but Xdebug isn't behaving for some reason, need to investigate...

dantleech: https://github.com/phpbench/phpbench/releases/tag/1.1.2

the tests worked locally with 8.1+ xdebug, so disabled them in 8.1/GH actions for now

PR #932, opened 2021-10-31
Do not sort the XAxes

### LINKED ISSUE #931: BUG: Mixed up labels in HTML output: bar_chart_time / benchmark
Got following HTML output (pay Attention to the small labels like `bench_parser_old_big_ast`):

image

Expected labels in the following order or as shown in the console:

bench_parser_old_big_ast
bench_parser_new_big_ast
bench_parser_old_medium_ast
bench_parser_new_medium_ast
bench_parser_old_small_ast
bench_parser_new_small_ast

image

The order of the test results is right but the labels seem to be mixed up in the HTML view.

dantleech:
Fixes #931

### DIFF
--- a/lib/Report/Model/BarChart.php
+++ b/lib/Report/Model/BarChart.php
@@ -67,12 +67,12 @@ class BarChart implements ComponentInterface
             return [];
         }
 
+        // use all unique discreet points as the X axes values
         $xAxes = array_unique(array_merge(...array_map(function (BarChartDataSet $dataSet) {
             return $dataSet->xSeries();
         }, $this->dataSets)));
-        sort($xAxes);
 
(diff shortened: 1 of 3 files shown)


lib/Report/Model/BarChart.php
```diff
@@ -67,12 +67,12 @@ public function xAxes(): array
             return [];
         }
 
+        // use all unique discreet points as the X axes values
         $xAxes = array_unique(array_merge(...array_map(function (BarChartDataSet $dataSet) {
             return $dataSet->xSeries();
         }, $this->dataSets)));
-        sort($xAxes);
```
dantleech, line 73: Not really sure why this was added :roll_eyes:

PR #940, opened 2021-11-06
Filter reports in addition to the run

dantleech:
Support filtering reports. Filters that apply to a bechmark run will also be applied to any other suite which is used as as a reference.

Additionally it adds a full set of benchmark "model" test builders to avoid having to use mocks.

### DIFF
--- a/lib/Console/Command/Handler/SuiteCollectionHandler.php
+++ b/lib/Console/Command/Handler/SuiteCollectionHandler.php
@@ -60,6 +60,9 @@ class SuiteCollectionHandler
         assert(is_array($files));
         assert(is_array($refs));
 
+        $subjectPatterns = $input->hasOption('filter') ? $input->getOption('filter') : [];
+        $variantPatterns = $input->hasOption('variant') ? $input->getOption('variant') : [];
+
         if (!$files && !$refs) {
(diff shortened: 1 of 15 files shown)


lib/Console/Command/Handler/SuiteCollectionHandler.php
```diff
@@ -60,6 +60,9 @@ public function suiteCollectionFromInput(InputInterface $input): SuiteCollection
         assert(is_array($files));
         assert(is_array($refs));
 
+        $subjectPatterns = $input->hasOption('filter') ? $input->getOption('filter') : [];
+        $variantPatterns = $input->hasOption('variant') ? $input->getOption('variant') : [];
```
dantleech, line 64: can add these options to the `report` command too.

lib/Model/Benchmark.php
```diff
@@ -45,6 +46,27 @@ public function __construct(Suite $suite, string $class)
         $this->class = $class;
     }
 
+    /**
+     * @param string[] $patterns
+     */
+    public static function matchesPatterns(string $benchmark, string $subject, array $patterns): bool
```
dantleech, line 52: move this somewhere

PR #942, opened 2021-11-16
Make it easy to copy composer install command on github

loevgaard:
This will allow users to click the 'copy to clipboard' button in the github interface


dantleech:
Thanks!

PR #948, opened 2021-11-21
Program executor

dantleech:
This PR experiments with adding a new executor which can execute benchmark variants according to "programs".

A program is a nestable sequence of "code snippets" which are used to composer the benchmarking script.

A program is a DSL which determines the execution flow, e.g.

```
warmup;
hrtime_sampler{
    memory_sampler{
        revsIteration{
            beforeMethods;
            callSubject;
            afterMethods;
        }
    }
}
```

Where each "stage" of the program is mapped to a class which can generate the code


lib/Executor/Parser/StageLexer.php
```diff
@@ -0,0 +1,85 @@
+<?php
+
+namespace PhpBench\Executor\Parser;
+
+use PhpBench\Expression\Token;
+use PhpBench\Expression\Tokens;
+
+final class StageLexer
```
dantleech, line 8: if we need a lexer, may as well abstract the one we have.

lib/Executor/ScriptExecutor.php
```diff
@@ -0,0 +1,118 @@
...
+    ) {
+        $this->scriptPath = $scriptPath;
+        $this->factory = $factory;
+        $this->scriptRemove = $scriptRemove;
+    }
+
+    /**
+     * @return array<string,mixed>
+     */
+    public function execute(/** include the run ID here **/string $script): array
```
dantleech, line 40: the idea was to include the run ID so that it can be used to namespace the file rather than use a `uniqid`.

lib/Expression/NodeEvaluator/DisplayAsEvaluator.php
```diff
@@ -22,7 +22,11 @@ public function evaluate(Evaluator $evaluator, Node $node, array $params): ?Node
             return null;
         }
 
-        $value = $evaluator->evaluateType($node->node(), NumberNode::class, $params);
+        $value = $evaluator->evaluate($node->node(), $params);
+
+        if ($value instanceof NullNode) {
```
dantleech, line 27: revert

lib/Model/Result/HrTimeResultFactory.php
```diff
@@ -0,0 +1,14 @@
...
+namespace PhpBench\Model\Result;
+
+use PhpBench\Model\ResultFactory;
+use PhpBench\Model\ResultInterface;
+
+final class HrTimeResultFactory implements ResultFactory
+{
+    public function create(array $data): ResultInterface
+    {
+        return new TimeResult(($data['net']), $data['revs'], 'nanoseconds');
```
dantleech, line 12: use TimeUnit constant

lib/Model/Result/TimeResult.php
```diff
@@ -54,11 +61,13 @@ public static function fromArray(array $values): ResultInterface
...
 
     /**
+     * TODO: This is downcasting to an int
+     *
      * Return the net-time for this iteration.
      */
     public function getNet(): int
     {
-        return $this->netTime;
+        return (int)$this->netTime;
```
dantleech, line 70: should be fine to change this to a float?

tests/Unit/Executor/Parser/StageParserTest.php
```diff
@@ -0,0 +1,94 @@
...
+        yield [
+            'stage;two{three}four',
+            new StageNode('root', [
+                new StageNode('stage'),
+                new StageNode('two', [
+                    new StageNode('three'),
+                ]),
+                new StageNode('four'),
+            ])
+        ];
```
dantleech, line 76: add case for `stage;two{three;four}five{six;seven};eight` or similar

lib/Executor/PhpProcessOptions.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace PhpBench\Executor;
+
+class PhpProcessOptions
```
dantleech, line 5: final

lib/Executor/ScriptBuilder.php
```diff
@@ -0,0 +1,76 @@
...
+                'Unknown stage "%s", known stages: "%s"',
+                $node->name,
+                implode('", "', array_keys($this->scriptStages))
+            ));
+        }
+
+        $stage = $this->scriptStages[$node->name];
+        $lines = $this->indent(array_merge([
+            sprintf('// >>> %s', $node->name),
+        ], $stage->start($context)), $indentation);
```
dantleech, line 47: s/node/unit

lib/Expression/Token.php
```diff
@@ -41,6 +41,10 @@ final class Token
     public const T_STRING = 'string';
     public const T_PARAMETER = 'parameter';
     public const T_QUESTION = 'question';
+    public const T_SEMICOLON = 'semicolon';
+    public const T_OPEN_BRACE = 'open_brace';
+    public const T_CLOSE_BRACE = 'close_brace';
```
dantleech, line 46: can remove these now?

lib/Model/Result/HrTimeResultFactory.php
dantleech: validate for array key existence

PR #949, opened 2021-11-30
Allow Symfony 6

julien-boudry:


dantleech:
Thanks, would fix #946

dantleech:
Replaced by #952

PR #951, opened 2021-12-01
Added simple check for glob paths in config.path value

### LINKED ISSUE #950: Support glob in runner.path
It would be great to be able to use a wildcard in `runner.path` to test in multiple directories without having to list them individually.

This could look like:

```json
{
    "$schema": "./vendor/phpbench/phpbench/phpbench.schema.json",
    "runner.bootstrap": "vendor/autoload.php",
    "runner.path": "modules/*/test/Benchmark",
}
```

And it would run in `modules/A/test/Benchmark`, `modules/B/test/Benchmark`, etc.

ricardoboss:
Fixes #950

### DIFF
--- a/lib/Extension/RunnerExtension.php
+++ b/lib/Extension/RunnerExtension.php
@@ -518,16 +518,36 @@ class RunnerExtension implements ExtensionInterface
             return;
         }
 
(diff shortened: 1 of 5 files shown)


dantleech:
thanks!

PR #953, opened 2021-12-04
Update CHANGELOG and refactor glob paths

dantleech:


lib/Util/PathNormalizer.php
```diff
@@ -0,0 +1,37 @@
...
+                if (empty($globPaths)) {
+                    return [];
+                }
+
+                return $globPaths;
+            }
+
+            return [ $path ];
+        }, $paths));
+    }
```
dantleech, line 31: @ricardoboss i extracted the logic to a static funtion and added dedicated tests. The original version didn't work with absolute paths

PR #972, opened 2021-12-28
Tell composer that the master branch is 1.2-dev

derrabus:
Currently, Packagist and Composer believe that the master branch contains version 1.1-dev. But since that branch is ahead of the 1.2.x releases, that seems to be a lie. 🙂

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -71,7 +71,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "1.1-dev"
+            "dev-master": "1.2-dev"
         }
     },
     "autoload": {


dantleech:
oops, tbh we could remove the branch alias. I don't think it's really useful?

derrabus:
It is useful if another package for some reason wants to use a dev version of your package.

Alternatively, we could also set it to `1.x-dev` which bascially tells Composer that the branch is ahead of all one-dot-something tags. In that case, you would only need to bump it when you're working on a new major version.

dantleech:
I think in that case they could use `dev-master` though.

PR #977, opened 2022-01-03
Allow float value instead of int

agustingomes:
There are parts of the application that would send float values to
this method.

This change intends to address a PHP Deprecation in PHP 8.1.


dantleech:
Thanks :+1:

Issue #978, opened 2022-01-07
Avoid using webmozart/path-util


ricardoboss:
Apparently, [webmozart/path-util](https://packagist.org/packages/webmozart/path-util) is no longer maintained and phpbench has it as a dependency.

You should consider changing it to a maintained package or the proposed one: [symfony/filesystem](https://packagist.org/packages/symfony/filesystem).

PR #981, opened 2022-02-17
Remove webmozart/path-util

Arzaroth:
Currently, composer complains when installing phpbench, with the following message :
~~~
Package webmozart/path-util is abandoned, you should avoid using it. Use symfony/filesystem instead.
~~~

This problem was brought up in issue #978

Indeed, the symfony/filesystem package provides the same functionalities with the class `Symfony\Component\Filesystem\Path`

As a caveat, it is only available since v5.4 so it could be considered a breaking change for some end users.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,11 +24,10 @@
         "psr/log": "^1.1 || ^2.0 || ^3.0",
         "seld/jsonlint": "^1.1",
         "symfony/console": "^4.2 || ^5.0  || ^6.0",
-        "symfony/filesystem": "^4.2 || ^5.0 || ^6.0",
+        "symfony/filesystem": "^5.4 || ^6.0",
         "symfony/finder": "^4.2 || ^5.0 || ^6.0",
         "symfony/options-resolver": "^4.2 || ^5.0 || ^6.0",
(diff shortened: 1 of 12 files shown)


dantleech:
Hm, kindof annoying that it's merged into a package which to which it has only a conceptual relation :/ I guess this means that anybody using SF 5.4 wouldn't be able to upgrade PHPBench though.

simPod (approved):

Arzaroth:
Although this was approved recently, I believe the discussion is still open.

@dantleech Another alternative would be to not rely symfony/filesystem at all and implement needed functions, in the package or another (thus continuing the legacy of webmozart/path-util in a way).

I do agree with the general feeling that it is a bit much to require the whole filesystem package only to manipulate paths.

dantleech:
i'm kinda tempted to fork webmozart/path, equally we do already depend on SF components and it's could fair to bump the min dependency now, esp. if that coincides with the 7.4 EOL

dantleech:
for now I have vendored the class and removd the webmozart dep. in the future we could switch to the "symfony" version when the dependency constriants allow it.

Issue #997, opened 2022-07-18
Missing Attribute::IS_REPEATABLE for Assert Attribute class


nyamsprod:
https://github.com/phpbench/phpbench/blob/facdd1259535ce70f76a5220ebe31476c96c4068/lib/Attributes/Assert.php#L7

I am getting the following error when using attributes

```
Attribute class PhpBench\Attributes\Assert is not repeatable but is already present above the method.
```

The code below produces the error
```php
    #[Bench\OutputTimeUnit('seconds')]
    #[Bench\Assert('mode(variant.mem.peak) < 2097152')]
    #[Bench\Assert('mode(variant.time.avg) < 10000000')]
    public function benchWithAttributes(): void
    {
        // internal bench code
    }
```

The same code using docblock works fine without error

```php
    /**
     * @OutputTimeUnit("seconds")]
     * @Assert("mode(variant.mem.peak) < 2097152")
     * @Assert("mode(variant.time.avg) < 10000000")
     */
    public function benchWithAnnotations(): void
    {
        //internal bench code
    }
```

dantleech:
happy to accept a PR

nyamsprod:
@dantleech PR submitted 

I am only testing that the repeatable issue is fixed and I assume the rest is correctly handled by the library.

Also as it was my first time using the library maybe other attributes may require the same fix but I did not check them. Maybe a future fix can be applied to them too.

dantleech:
possibly it's required on others too, but for the meantime tagged 1.2.6 :)

PR #998, opened 2022-07-18
#997 make Assert repeatable for PHP8+ attributes

nyamsprod:
Fix for missing `Attribute::IS_REPEATABLE` for the `Assert` attribute see #997

### DIFF
--- a/lib/Attributes/Assert.php
+++ b/lib/Attributes/Assert.php
@@ -4,7 +4,7 @@ namespace PhpBench\Attributes;
 
 use Attribute;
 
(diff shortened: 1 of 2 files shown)


nyamsprod:
Seems the tests are OK only documentation build is failing but seems to be a recurring issue with other PRs as well.

dantleech:
yes, i need to fix that documentation issue, but also don't want to make the main branch red. i will try and fix the docs later today.

dantleech:
ok, hopefully the documentation build is fixed now, thanks :)

PR #1000, opened 2022-07-24
Replace webmozart/path-util with symfony/filesystem

mamazu:


dantleech:
Haven't merged the same PR https://github.com/phpbench/phpbench/pull/981 due to it preventing people from using SF 4.4 from using PHPBench.

mamazu:
Then I will close this one and we just use the other one? I mean we could just use the > constraint then we should be save with all new symfony versions. But I agree this totally doesn't make sense.

dantleech:
Am vendoring the class instead.

PR #1002, opened 2022-07-30
GH-1001: Fix bad exception call

dantleech:
Fix issue fixed in PR #1001 in a slightly different way and adds a test


TRowbotham:
My only suggestion might be to store the `array_keys()` call in a variable, since the function is called twice as it may be expensive on larger parameter sets. Otherwise, it looks good. Thanks!

dantleech:
done, thanks :)

PR #1009, opened 2022-10-15
Webmozart/path vendoring

### LINKED ISSUE #978: Avoid using webmozart/path-util

dantleech:
I've decided to simply copy the `symofny/filesystem` Path class to PHPBench. Perhaps in the future we can switch to using the Filesystem version when the version constraints allow it - even if that package has nothing to do with this class...

Fixes #978

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ master
 
 Improvements:
 
+- Remove dependency on `webmozaty/path` (copied the class into PHPBench)
 - Update PHPStan to 1.0 and fix new errors #1006
 - Use `webmozart/glob` instead of `glob` for benchmark paths and config
   include paths #1005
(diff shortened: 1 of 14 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ master
 
 Improvements:
 
+- Remove dependency on `webmozaty/path` (copied the class into PHPBench)
```
ricardoboss, line 9: (nitpick) Typo

dantleech: thanks will fix in following PR

PR #1011, opened 2022-10-24
Detect xdebug.use_compression and set correct extension

blackwolf12333:
Xdebug 3.1 enables compression of the cachegrind files automatically so the extension of the output by default is not .cachegrind but .cachegrind.gz.

This patch adds a check to see if compression is enabled and then uses .cachegrind.gz instead of .cachegrind as extension.

### DIFF
--- a/extensions/xdebug/lib/XDebugExtension.php
+++ b/extensions/xdebug/lib/XDebugExtension.php
@@ -44,10 +44,14 @@ class XDebugExtension implements ExtensionInterface
 
     public function load(Container $container): void
     {
+        $container->register(XDebugUtil::class, function (Container $container) {
+            return XDebugUtil::fromEnvironment();
+        });
         $container->register(ProfileCommand::class, function (Container $container) {
             return new ProfileCommand(
                 $container->get(RunnerHandler::class),
-                $container->get(self::PARAM_OUTPUT_DIR)
+                $container->get(self::PARAM_OUTPUT_DIR),
+                $container->get(XDebugUtil::class)
             );
         }, [
             ConsoleExtension::TAG_CONSOLE_COMMAND => []
@@ -64,6 +68,7 @@ class XDebugExtension implements ExtensionInterface
             return new CompositeExecutor(
                 new ProfileExecutor(
                     $container->get(RemoteExecutor::class),
+                    $container->get(XDebugUtil::class),
                     $container->getParameter(CoreExtension::PARAM_WORKING_DIR)
                 ),
                 $container->get(RemoteMethodExecutor::class)
--- a/extensions/xdebug/lib/XDebugUtil.php
+++ b/extensions/xdebug/lib/XDebugUtil.php
@@ -13,10 +13,37 @@
 namespace PhpBench\Extensions\XDebug;
 
 use PhpBench\Executor\ExecutionContext;
+use RuntimeException;
 
 class XDebugUtil
 {
-    public static function filenameFromContext(ExecutionContext $context, $extension = ''): string
+    public static function fromEnvironment(): XDebugUtil
+    {
+        $xdebugVersion = phpversion('xdebug');
+        $xdebugUseCompression = ini_get('xdebug.use_compression') === '1';
+
+        return new XDebugUtil($xdebugVersion, $xdebugUseCompression);
+    }
+
+    /**
+     * @var string|false
+     */
+    private $xdebugVersion;
+    /**
+     * @var bool
+     */
+    private $xdebugUseCompression;
+
+    /**
+     * @param string|false     $xdebugVersion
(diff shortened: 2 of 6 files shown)


dantleech:
thanks, looks ok. would be nice to have a test, maybe inject the "xdebug version" from the container? also need to run CS fixer, there is a `composer integrate` command which should run all the tooling locally.

blackwolf12333:
I'll take a look!

blackwolf12333:
Okay I got the tests running, and actually found a small issue in the implementation which I've now corrected.
I'm still not really sure how to write a test for this, since I can't really inject a different xdebug version. Do you have any suggestions?

dantleech:
>  Do you have any suggestions? Because I'd really prefer to have a test for this myself.

basically inject the xdebug version fro mthe container, then you only need to test the class with the XDebug version (e.g. `new XdebugUtil('3.1.2');`?

extensions/xdebug/lib/XDebugExtension.php
```diff
@@ -44,10 +44,14 @@ public function configure(OptionsResolver $resolver): void
 
     public function load(Container $container): void
     {
+        $container->register(XDebugUtil::class, function (Container $container) {
+            return new XDebugUtil();
```
dantleech, line 48: `new XDebugUtil(phpversion('xdebug'))` ?

dantleech (approved):

blackwolf12333:
Fixed!

extensions/xdebug/lib/XDebugUtil.php
```diff
@@ -13,10 +13,23 @@
...
+use RuntimeException;
 
 class XDebugUtil
 {
-    public static function filenameFromContext(ExecutionContext $context, $extension = ''): string
+    /**
+     * @var string|false
+     */
+    private $xdebugVersion;
+    private $xdebugUseCompression;
```
dantleech, line 24: missing type hint

extensions/xdebug/lib/XDebugUtil.php
```diff
@@ -13,10 +13,23 @@
...
+
+    public function __construct($xdebugVersion, $xdebugUseCompression)
```
dantleech, line 26: missing type hints @param and whatever `$xdebugUseCompression` is, I think this directory is missing from the PHPStan paths.

extensions/xdebug/lib/XDebugExtension.php
```diff
@@ -44,10 +44,14 @@ public function configure(OptionsResolver $resolver): void
 
     public function load(Container $container): void
     {
+        $container->register(XDebugUtil::class, function (Container $container) {
+            return new XDebugUtil(phpversion('xdebug'), ini_get('xdebug.use_compression'));
```
dantleech, line 48: prob need to cast `ini_get` to a bool?

dantleech: probably need to handle these values properly - what does `phpversion` return if `xdebug` is not set? `init_get` will return `false` if the init settings does not exist, but a string otherwise... does it return `1` and `0` ?

extensions/xdebug/lib/XDebugUtil.php
dantleech: ouch, I really need to bump the PHP version to 7.4 at least :D

extensions/xdebug/lib/XDebugUtil.php
```diff
@@ -13,10 +13,41 @@
...
 class XDebugUtil
 {
-    public static function filenameFromContext(ExecutionContext $context, $extension = ''): string
+    public static function fromEnvironment(): XDebugUtil
+    {
+        $xdebugVersion = phpversion('xdebug');
+
+        if ($xdebugVersion === false) {
+            throw new RuntimeException(
+                'Xdebug is not installed'
```
dantleech, line 26: image

this is eagery loaded, so we do indeed need to accept `false|string` I think.

extensions/xdebug/lib/XDebugUtil.php
```diff
@@ -13,10 +13,34 @@
...
+    /**
+     * @var string|false
+     */
+    private $xdebugVersion;
+    /**
+     * @var bool
+     */
+    private $xdebugUseCompression;
+
+    public function __construct(string $xdebugVersion, bool $xdebugUseCompression)
```
dantleech, line 37: not a string, can remove the type hint and add a docblock:

```
/** @param string|false $xdebugVersion */
```

dantleech:
just one last comment....

dantleech (approved):

dantleech:
:boom: and updated the changelog, thanks

PR #1013, opened 2022-12-19
Add support for annotations 2

greg0ire:
It should allow other packages to add support for `doctrine/lexer` 3

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "ext-reflection": "*",
         "ext-spl": "*",
         "ext-tokenizer": "*",
-        "doctrine/annotations": "^1.13",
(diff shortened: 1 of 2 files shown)


greg0ire:
Obviously, I need to look into this more in depth :sweat_smile:

dantleech:
Might be worth dropping 7.3/4 support if this is tricky.

greg0ire:
Just noticed removing php-cs-fixer allows me to reproduce errors locally (it requires annotations as well, so that makes sense).

dantleech:
https://github.com/phpbench/phpbench/pull/1015/commits/eb95d026b49f334031f1e846c0e7b28c51958b37

seems to fix the build

dantleech:
Thanks @greg0ire

PR #1014, opened 2022-12-19
Address Package name change

greg0ire:
Using v3 seems to be the correct move: https://github.com/sphinx-doc/sphinx/blob/v5.3.0/Makefile#L1

This small PR should allow me to achieve non-first-contributor status, and no longer require you pressing a button for me to be able to run workflows.


dantleech:
:bow:

PR #1020, opened 2023-02-14
If \PhpBench\Executor\Method\ErrorHandlingExecutorDecorator throws unhandled exception now is a bit clearer why

MekDrop:
3rd exception parameter $throwable can be used to get original exception. 

I think it could be useful when debugging exceptions.


dantleech (approved):

dantleech:
yep, good call

MekDrop:
@dantleech any idea why tests are failing? I tried to look into but I can't say that I see something there related to my changes :| Should I do something?

dantleech:
hm, no it seems there are no failures just warnings? not sure why that behvior would have changed.

dantleech:
sorry this got lost, thanks!

PR #1029, opened 2023-07-18
Add Composer keywords for findability on Packagist

TravisCarden:
I almost didn't find you. 🙂 Composer keywords appear as filterable tags on Packagist, e.g., [#performance](https://packagist.org/search/?tags=performance).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,13 @@
 {
     "name": "phpbench/phpbench",
     "description": "PHP Benchmarking Framework",
+    "keywords": [
+        "benchmarking",
+        "optimization",
+        "performance",
+        "profiling",
+        "testing"
+    ],
     "license": "MIT",
     "authors": [
         {


dantleech:
Great, thanks!

TravisCarden:
[Nice](https://packagist.org/packages/phpbench/phpbench#dev-master): 🙂

phpbench-packagist-tags


FYI: A coworker pointed out to me that the `testing` keyword has a (desirable) side effect. (Thanks,@danepowell!)  Per https://getcomposer.org/doc/04-schema.md#keywords:

> Note: Some special keywords trigger composer require without the `--dev` option to prompt users if they would like to add these packages to require-dev instead of require. These are: `dev`, `testing`, `static analysis`.

Example:

```
$ composer require phpbench/phpbench
The package you required is recommended to be placed in require-dev (because it is tagged as "testing") but you did not use --dev.
Do you want to re-run the command with --dev? [yes]?
```

c.f. https://github.com/composer/composer/commit/d2f37b5680858435fc292e2a0502deb4a33c2d06#diff-5cba376e8fa1a6588af9c27e0a72545536660bf3f6083292ea8f59f28890926cR249

Happy accidents, huh? 😛

PR #1032, opened 2023-09-02
Correct phpbench.schema.json

johnbillion:
These two config options are arrays, not objects.

I think several of the other config options defined as objects are also arrays, but I've not used or tested them to be sure.

### DIFF
--- a/phpbench.schema.json
+++ b/phpbench.schema.json
@@ -83,13 +83,13 @@
         "runner.env_enabled_providers": {
             "description": "Select which environment samplers to use",
             "type": [
-                "object"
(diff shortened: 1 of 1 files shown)


dantleech:
The schema is generated, so will need to fix the generator rather

szepeviktor:
> need to fix the generator

https://github.com/phpbench/phpbench/blob/6d269ebfcbf8ea44d7ffe2a4deb9f2509027f74c/lib/Development/ConfigSchemaDumper.php#L84-L89

dantleech:
closing for now.

PR #1034, opened 2023-09-18
Add documentation about adding an environment provider

GromNaN:
I worked on adding an [environment provider for MongoDB benchmarks](https://github.com/mongodb/mongo-php-library/blob/4e581cf0c5f9a901b189f1b247a3a7759be18d02/benchmark/Extension/EnvironmentProvider.php) and found some missing pieces of documentation.

This needs to be reviewed because I just started working with PHPBench. But I think this new documentation will be helpful if someone needs to create a custom env provider.

Also, I added the namespaces to the base extension example. I was disturbed to not knowing the actual class of `ExtensionInterface`.

### DIFF
--- a/docs/customisation.rst
+++ b/docs/customisation.rst
@@ -8,4 +8,5 @@ Customisation
    customisation/command
    customisation/executor
    customisation/progress-logger
+   customisation/environment-provider
    customisation/report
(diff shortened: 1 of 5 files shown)


dantleech:
Sorry, this got lost in my inbox - looks good to me, are you happy for it to be merged?

dantleech:
too late, thanks :slightly_smiling_face:

PR #1039, opened 2023-10-30
Fix sf 7 compat

dantleech:


examples/Extension/AcmeExtension.php
```diff
@@ -83,7 +83,7 @@ public function load(Container $container): void
         // endsection: executor_di
 
         // section: env_provider_di
-        $container->register(AcmeExecutor::class, function (Container $container) {
+        $container->register(HomeProvider::class, function (Container $container) {
```
dantleech, line 86: not really sure how this passed CI but this was using the wrong service name cc @keulinho

keulinho:
Thanks for fixing! :+1: I was relying on the CI to catch stuff like that

PR #1040, opened 2023-11-05
Chore: Allow `runner.path` to accept array in schema.json.

KennedyTedesco:


dantleech:
see https://github.com/phpbench/phpbench/pull/1032 :)

PR #1049, opened 2023-11-30
Add ergebnis/composer-normalize package

aivchen:
Would be cool to keep composer.json ordered:)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -38,6 +38,43 @@ jobs:
                 name: "Validate composer.json"
                 run: "composer validate --strict --no-check-lock"
 
+    composer-normalize:
+        needs:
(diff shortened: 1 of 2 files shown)


dantleech:
needs a rebase now -- but not sure it's worth putting it in CI -- maybe just add it to the `integrate` target in `composer.json` - that way it will be _eventually_ ordered :smile: but either way I'm happy.

aivchen:
@dantleech Rebased!

>maybe just add it to the integrate target in composer.json

It's already there:)

> that way it will be eventually ordered

Well, in my opinion it must be in CI. For example, can we exclude php-cs-fixer from CI in the same way?:blush:

dantleech:
Well, I'd argue it's less important, but ok!

PR #1052, opened 2023-12-03
Improved types in tests (preparing for PHPStan level 6)

aivchen:
On phpstan level 6 there are around 700 errors in the project. I think it will be better to fix them in batches. What do you think?

### DIFF
--- a/lib/Registry/Config.php
+++ b/lib/Registry/Config.php
@@ -29,8 +29,13 @@ class Config extends ArrayObject
      */
     final public const NAME_REGEX = '{^[0-9a-zA-Z_-]+$}';
 
+    /** @var string */
     private $name;
 
+    /**
+     * @param string $name
+     * @param mixed[] $config
+     */
     public function __construct($name, array $config)
     {
         if (!preg_match(self::NAME_REGEX, (string) $name)) {
@@ -57,6 +62,9 @@ class Config extends ArrayObject
         return parent::offsetGet($offset);
     }
 
+    /**
+     * @return string
+     */
     public function getName()
     {
         return $this->name;
--- a/lib/Registry/ConfigurableRegistry.php
+++ b/lib/Registry/ConfigurableRegistry.php
@@ -63,7 +63,7 @@ class ConfigurableRegistry extends Registry
     /**
      * Return the named configuration.
      *
-     * @param string|array $name
+     * @param string|mixed[] $name
      */
     public function getConfig($name): Config
     {
@@ -112,6 +112,8 @@ class ConfigurableRegistry extends Registry
      *
      * Note that all configurations must be associated with a named service
      * via a configuration key equal to the configuration service type of this registry.
+     *
+     * @param mixed[] $config
      *
      */
     public function setConfig(string $name, array $config): void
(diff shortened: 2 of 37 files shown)


dantleech:
Hmm, should probably merge #1047 before going too far down this path

aivchen:
Oh, ok. Let's wait:)

lib/Registry/Config.php
```diff
@@ -29,8 +29,13 @@ class Config extends ArrayObject
      */
     final public const NAME_REGEX = '{^[0-9a-zA-Z_-]+$}';
 
+    /** @var string */
     private $name;
 
+    /**
+     * @param string $name
+     * @param mixed[] $config
```
dantleech, line 37: not important but i guess this is the same as `array<array-key,mixed>`, but it really should always be `array<string,mixed>` IIRC

lib/Registry/ConfigurableRegistry.php
```diff
@@ -63,7 +63,7 @@ public function __construct(
     /**
      * Return the named configuration.
      *
-     * @param string|array $name
+     * @param string|mixed[] $name
```
dantleech, line 66: should be `string[]`

dantleech, line 66: should be `string|array<string,mixed>` (i.e. the `config` record)

lib/Registry/ConfigurableRegistry.php
```diff
@@ -113,6 +113,8 @@ public function getConfigNames(): array
      * Note that all configurations must be associated with a named service
      * via a configuration key equal to the configuration service type of this registry.
      *
+     * @param mixed[] $config
```
dantleech, line 116: array<string,mixed>

tests/Benchmark/Console/Command/RunCommandBench.php
```diff
@@ -63,6 +63,10 @@ class NothingBench { public function benchNothing(): void {}}
         );
     }
 
+    /**
+     * @param mixed[] $args
+     * @param mixed[] $config
```
dantleech, line 68: $args => array<string,mixed>` config also

tests/Benchmark/IntegrationBenchCase.php
```diff
@@ -19,6 +19,9 @@ protected function workspace(): Workspace
         return Workspace::create(__DIR__ . '/../Workspace');
     }
 
+    /**
+     * @param mixed[] $config
+     */
     protected function container(array $config = []): ContainerInterface
```
dantleech, line 25: ditto

dantleech:
looks good. almost all the `$config` things should be `array<string,mixed>` (i.e. a "record) and a few other things are too broadly typed, but not sure it would pay too be too strict at this point in time. if we can make phpstan happy at level 6 it's already good :slightly_frowning_face:

aivchen:
> looks good. almost all the `$config` things should be `array<string,mixed>`

You didn't give me a chance to fix it:))

>if we can make phpstan happy at level 6 it's already good

Yup, at level 6 we still can make some tricks. But at higher levels we will have to get rid of the mixed type.

PR #1053, opened 2023-12-05

aivchen:


lib/Benchmark/SamplerManager.php
```diff
@@ -55,7 +56,7 @@ public function addSamplerCallable(string $name, $callable): void
             throw new InvalidArgumentException(sprintf(
                 'Given sampler "%s" callable "%s" is not callable.',
                 $name,
-                is_string($callable) ? $callable : gettype($callable)
+                is_string($callable) ? $callable : gettype($callable) // @phpstan-ignore-line
```
dantleech, line 59: no need to change, but just for information I found out about [get_debug_type](https://www.php.net/manual/en/function.get-debug-type.php) which is better than `gettype` here.

lib/Config/ConfigProcessor.php
```diff
@@ -4,5 +4,10 @@
 
 interface ConfigProcessor
 {
+    /**
+     * @param array<string, mixed> $config
+     *
+     * @return array<string, mixed>
```
dantleech, line 10: i wonder if it would be worth adding a type alias for "config" records:

https://phpstan.org/writing-php-code/phpdoc-types#local-type-aliases

so this could then be `@return config-record` with `@phpstan-import-type config-record from Config` in the class docblock or similar

dantleech (approved):

dantleech:
great work :slightly_smiling_face:

lib/Benchmark/SamplerManager.php
aivchen: Yes, `get_debug_type` is cool. But I think it breaks B/C as some users can rely on Exception message. So it can't be included into minor/patch release.
With `get_debug_type` the message will look like this if `$callable === 'bar'`: `Given sampler "foo" callable "string" is not callable.`
And with gettype the message looks like this: `Given sampler "foo" callable "bar" is not callable.`

lib/Config/ConfigProcessor.php
aivchen: Great idea! I'll try to implement it in next PR:)

lib/Benchmark/SamplerManager.php
dantleech: i hope users don't rely on the exception message, but valid point :)

PR #1055, opened 2023-12-06
improved types (PHPStan level 6)

aivchen:
All tests are covered!

Regarding https://github.com/phpbench/phpbench/pull/1053#discussion_r1417008223

To be honest, it is not clear to me where we need to use the `config-record`, because I do not understand whether this or that part of the code is related to `Config` or not. Therefore, I would leave everything as it is for now. 

By the way, what is the global alias `parameters` responsible for? It has `array<string, mixed>` type and looks like a `config-record` you wanted to add:)


dantleech:
Yeah, I think "parameters" was for this purpose indeed. Fine to leave it for now, thanks!

PR #1057, opened 2023-12-06
Set strict indent for composer.json

aivchen:


dantleech:
thanks again :slightly_smiling_face:

PR #1059, opened 2023-12-08

aivchen:


lib/Model/Benchmark.php
```diff
@@ -110,7 +106,7 @@ public function getSuite(): Suite
     }
 
     /**
-     * @return ArrayIterator<Subject>
+     * @return ArrayIterator<array-key, Subject>
```
dantleech, line 109: is the `ArrayIterator<Subject>` form not valid? but happy with the more explict form in anycase

lib/Model/Error.php
```diff
@@ -20,6 +20,14 @@
...
 {
+    /**
+     * @param string $message
+     * @param string $class
+     * @param int $code
+     * @param string $file
+     * @param int $line
+     * @param string $trace
+     */
     public function __construct(private $message, private $class, private $code, private $file, private $line, private $trace)
```
dantleech, line 31: any reason not to add real type hints here?

lib/Model/Error.php
```diff
@@ -36,31 +44,49 @@ public static function fromException(Throwable $exception): Error
...
+     */
     public function getClass()
     {
         return $this->class;
     }
 
+    /**
+     * @return string
+     */
     public function getTrace()
```
dantleech, line 90: ditto, seems these can all be real return types?

lib/Model/ResultCollection.php
```diff
@@ -106,7 +113,12 @@ public function getMetric(string $class, string $metric)
...
      * clearly a problem and we should allow an error to be thrown.
      *
+     * @template TDefault
+     *
+     * @param class-string<ResultInterface> $class
+     * @param TDefault $default
      *
+     * @return int|float|TDefault
      */
     public function getMetricOrDefault(string $class, string $metric, $default = null)
```
dantleech, line 123: :+1:

lib/Model/Summary.php
```diff
@@ -26,6 +26,8 @@ class Summary
     private int $nbFailures = 0;
     private int $nbAssertions = 0;
     private int $nbErrors = 0;
+
+    /** @var array<string, array<int|float>> */
```
dantleech, line 30: _could_ use an array shape here, although maybe better as a type alias if it's referenced again

lib/Model/Summary.php
```diff
@@ -120,41 +122,65 @@ public function getNbAssertions(): int
         return $this->nbAssertions;
     }
 
+    /**
+     * @return array<string, array<int|float>>
+     */
     public function getStats(): array
```
dantleech, line 128: i.e. we could return that array shape here

lib/Model/Summary.php
```diff
@@ -120,41 +122,65 @@ public function getNbAssertions(): int
...
+     */
     public function getMeanStDev()
     {
         return Statistics::mean($this->stats['stdev']);
     }
 
+    /**
+     * @return int|float
+     */
     public function getMeanRelStDev()
```
dantleech, line 184: could use real RTs?

lib/Model/Variant.php
```diff
@@ -94,6 +93,7 @@ public function createIteration(array $results = []): Iteration
     /**
      * Return the iteration at the given index.
      *
+     * @param int $index
      */
     public function getIteration($index): ?Iteration
```
dantleech, line 98: ...

dantleech (commented):
nice work, I guess you opted to use docblocks rather than real types for B/C reasons? i wonder if we can add the real types and either:

1) bump to 2.0
2) assume that nobody has overridden this stuff (it's highly unlikely)
3) keep the docblocks and migrate them at a later date with rector.

happy to got with 3)

lib/Model/Benchmark.php
aivchen: Yes, it's not valid:

> Method PhpBench\Model\Benchmark::getIterator() return type with generic class ArrayIterator does not specify its types: TKey, TValue

lib/Model/Error.php
aivchen: B/C

lib/Model/Error.php
aivchen: B/C

lib/Model/Summary.php
aivchen: B/C :smile:

lib/Model/Variant.php
aivchen: B/C

aivchen:
> I guess you opted to use docblocks rather than real types for B/C reasons?

Exactly:)

>2. assume that nobody has overridden this stuff (it's highly unlikely)

It's unlikely but we can't exclude it:)

> 3. keep the docblocks and migrate them at a later date with rector.

Yes. I think we need to type eveything with dockblocks for now. After that we can prepare 2.0 and run rector. Also will be cool to finalize all classes in 2.0 to have more freedom in classes changing.

lib/Model/Summary.php
aivchen: Let me think a bit.

dantleech:
Sounds good!

PR #1062, opened 2023-12-12
Fixed TypeError in MemoryUnit.php

aivchen:
There are two methods in `lib/Util/MemoryUnit.php`:

https://github.com/phpbench/phpbench/blob/5f12699b84926d7674fc5ad54e319a7556d0fc50/lib/Util/MemoryUnit.php#L90-L94

and

https://github.com/phpbench/phpbench/blob/5f12699b84926d7674fc5ad54e319a7556d0fc50/lib/Util/MemoryUnit.php#L118-L122

`$unit` is nullable and can be returned as is. But return type of the methods is string. So if `$unit === null` we get `TypeError: PhpBench\Util\MemoryUnit::resolveSuitableBinaryUnit(): Return value must be of type string, null returned`

### DIFF
--- a/lib/Util/MemoryUnit.php
+++ b/lib/Util/MemoryUnit.php
@@ -85,9 +85,9 @@ class MemoryUnit
     }
 
     /**
-     * Resolve an binary unit
+     * Resolve a binary unit
      */
-    public static function resolveSuitableBinaryUnit(?string $unit, ?float $value): string
+    public static function resolveSuitableBinaryUnit(string $unit, ?float $value): string
     {
         if ($unit !== self::AUTO) {
             return $unit;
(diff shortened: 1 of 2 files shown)


dantleech:
I think NULL could be passed however: https://github.com/phpbench/phpbench/blob/914cee8e677165a547240c9da459b5914a063a3e/lib/Expression/NodePrinter/DisplayAsPrinter.php#L72

I don't think that can happen however, but according to the type system it could (value is mixed)

aivchen:
Look a bit above:
https://github.com/phpbench/phpbench/blob/914cee8e677165a547240c9da459b5914a063a3e/lib/Expression/NodePrinter/DisplayAsPrinter.php#L70

`TimeUnit::isTimeUnit` accepts only strings.

More:
https://github.com/phpbench/phpbench/blob/914cee8e677165a547240c9da459b5914a063a3e/lib/Expression/NodePrinter/DisplayAsPrinter.php#L40-L44
there is a check for `StringNode` which always returns a string from `StringNode::value()` method

>but according to the type system it could (value is mixed?)

I assume IDE hinted you that the value is mixed? In reality only string can be there. We just need to add  `@return string` to `StringNode::value()` and IDE will think it's a string:)

dantleech:
cool, thanks for checking :)

PR #1070, opened 2023-12-19
Phpstan stub for Psr\Container

aivchen:
This will let us know the type of the object returned by container:
```php
\PHPStan\dumpType($this->container()->get(SomeClass:class)) // "mixed" before change and "SomeClass" after the change
```


dantleech:
We could also add it directly to the Container package as the PSR ContainerInterface isn't used anyway. But happy to merge this

Issue #1069, opened 2023-12-19
RunnerConfig code inconsistency


aivchen:
Value read:
https://github.com/phpbench/phpbench/blob/0989c350d1a20e04b5ba6134d59b4ac938e84eb9/lib/Benchmark/RunnerConfig.php#L142-L148

`$default` param is null be default. But return type is array.

Value write:
https://github.com/phpbench/phpbench/blob/0989c350d1a20e04b5ba6134d59b4ac938e84eb9/lib/Benchmark/RunnerConfig.php#L286-L294

`$iterations` param is null by default. So the new instance can have `$iterations === null` property.

In my opinion it must look like this:
```php
    /**
     * Override the number of iterations to execute.
     */
    public function getIterations(array $default = []): array
    {
        return $this->iterations ?: $default;
    }

    public function withIterations(array $iterations = []): self
    {
        $this->assertArrayValuesGreaterThanZero($iterations);

        $new = clone $this;
        $new->iterations = $iterations;

        return $new;
    }
```
but it breaks B/C (or we can consider this solution as a bugfix:)))

What's the best way to solve the problem (if it's a problem at all) in your opinion?

aivchen:
I found a workaround in #1071

PR #1071, opened 2023-12-20
Improved types for RunnerConfig.php

aivchen:


dantleech:
Why not set the default as you originally suggested?

aivchen:
Because I found a solution not breaking B/C.

dantleech:
Is it a B/C break? Passing NULL would have resulted in a crash?

aivchen:
Seems I'm confused a bit. Let's see what we have.

Initial method (from master branch):

https://github.com/phpbench/phpbench/blob/a2c16caab52856f9e0bbe672856406ef3f8a0139/lib/Benchmark/RunnerConfig.php#L145-L148

And if I call `getIterations` without params I'll get TypeError as this method must return array:
```php
$myConfig = RunnerConfig::create();
$myConfig->getIterations(); // Uncaught TypeError: PhpBench\Benchmark\RunnerConfig::getIterations(): Return value must be of type array, null returned
```

In #1069 I suggested this solution:
```php
    public function getIterations(array $default = []): array
    {
        return $this->iterations ?: $default;
    }
```
But it's breaks B/C as some user can have the following code:
```php
$myConfig = RunnerConfig::create();
$myConfig->getIterations(null); // Uncaught TypeError: PhpBench\Benchmark\RunnerConfig::getIterations(): Argument #1 ($default) must be of type array, null given
```

So to keep B/C we must not change method signature. In this PR we just check the value of `$default` param. And if `$default === null` we set it to an empty array:
https://github.com/phpbench/phpbench/blob/603f3de496a1b8cf1c0bb64bcdea4816f9b6ad5e/lib/Benchmark/RunnerConfig.php#L107-L112

As you see the signature is not changed in comparing with master branch.

And this code will keep working:
```php
$myConfig = RunnerConfig::create();
$myConfig->getIterations();
$myConfig->getIterations(null);
```

On the other hand, users couldn't have the code like this as they would get TypeError. So in theory we can consider this case as a bug and use the solution from #1069. What do you think?

dantleech:
yes, I think B/C is only if they could depend on this behavor, in theory they could depend on the runtime error, but that would be truly horrifying. It's a bug fix :+1:

aivchen:
Fixed:)

dantleech (approved):

dantleech:
thanks!

PR #1075, opened 2023-12-28
Minor fix: Import `Attribute` symbol in `Format` and `RetryThreshold` Attributes

gsteel:


dantleech:
oops, was this not working ?!

gsteel:
It works fine @dantleech, but if you're using Psalm/PHPStan, they complain that the attributes aren't attributes!

dantleech:
somehow some CS issues need fixing

dantleech:
Thanks!

PR #1080, opened 2024-01-15
Forced `--no-ansi` in tests

aivchen:
Forced `--no-ansi` in tests as without this option it breaks tests if they are running from the terminal supporting colors.


aivchen:
Bump:)

dantleech:
Needs CS fixes

dantleech:
merci!

PR #1081, opened 2024-01-16

aivchen:
I had to disable 'void_return' php-cs-fixer rule. Otherwise it adds return type to `\PhpBench\Progress\LoggerInterface` after my changes.


aivchen:
Bump

lib/Progress/Logger/BlinkenLogger.php
```diff
@@ -177,9 +180,7 @@ private function drawIterations(Variant $variant, array $specials, $tag): void
         $line = sprintf('%-' . self::INDENT . 's', '#' . $variant->getSubject()->getIndex());
         $nbIterations = $variant->count();
 
-        for ($index = 0; $index < $nbIterations; $index++) {
-            $iteration = $variant->getIteration($index);
-
+        foreach ($variant as $index => $iteration) {
```
dantleech, line 183: Are we sure that index is a sequential list here? Maybe use array_values if not

aivchen: Yes, we are sure.
- In `lib/Model/Variant.php` we add values to the private property `Variant::$iterations` using only `$this->iterations[] = XXX`.
- `Variant::$iterations` has `/** @var list<Iteration> */` docblock. It means this array is 0-based sequential list (https://phpstan.org/writing-php-code/phpdoc-types#lists).

lib/Storage/HistoryEntry.php
```diff
@@ -20,10 +20,38 @@
...
+        private                   $tag,
+        private                   $branch,
+        private                   $nbSubjects,
+        private                   $nbIterations,
+        private                   $nbRevolutions,
+        private                   $minTime,
+        private                   $maxTime,
+        private                   $meanTime,
+        private                   $meanRelStDev,
+        private                   $totalTime
```
dantleech, line 48: :smile:

dantleech:
great work again :bow:

lib/Progress/Logger/BlinkenLogger.php
dantleech: yes, have some ancient memories about making for loops like that to work around "lists" that had been filtered etc making their array keys non-sequential.

PR #1091, opened 2024-04-09
Specify `--no-gpg-sign` in git integration tests

jbboehr:
`git` may use the user's global configuration which may specify `commit.gpgsign=true`


dantleech:
eh, looks like cs-fixer has broken since the last green build - any chance you would be able to check it?

you can run `composer integrate` to run most of the above checks including cs-fixer.

jbboehr:
Strange, it worked when I ran it yesterday. Maybe this: https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/pull/7923

dantleech:
eh, also weird that it worked on the other MR, anyway, thanks!

PR #1093, opened 2024-05-22
Allow php 8.2

jeremy379:
I had trouble installing the package on a PHP 8.2 environment. It seems the composer.json was restricted to 8.1

This PR made the update in composer.json and update the CI as well.


dantleech:
hm -- `^8.1` should allow any `8.x` _after_ `8.1`

image

https://semver.madewithlove.com/?package=php&constraint=%5E8.1

PR #1115, opened 2025-06-04
Fix typo error on customisation/extension documentation

### LINKED ISSUE #1114: Docs - Customization / Extensions -- typo error
I'm almost sure that a prefix is missing in **Customization / Extensions** page, following PHPBench JSON schema !

While we can read `core.extensions` at https://github.com/phpbench/phpbench/blob/master/docs/customisation/extension.rst?plain=1#L47

We only read `extensions` at https://github.com/phpbench/phpbench/blob/master/docs/customisation/extension.rst?plain=1#L79

Should be `core.extensions` at line 79 too.

llaville:
Fixes #1114

### DIFF
--- a/docs/customisation/extension.rst
+++ b/docs/customisation/extension.rst
@@ -76,7 +76,7 @@ You can activate and use your extension as follows ``phpbench.json``:
 
     {
         "$schema":"./vendor/phpbench/phpbench/phpbench.schema.json",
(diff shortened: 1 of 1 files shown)


dantleech:
sorry, upstream build was failing and now fixed, you'll need to rebase/update your MRs.

PR #1128, opened 2025-11-05
Use php-cs-fixer/shim instead + full Symfony 8 support

Kocal:
Follows #1125

### DIFF
--- a/lib/Extension/CoreExtension.php
+++ b/lib/Extension/CoreExtension.php
@@ -93,7 +93,7 @@ EOT
 
             foreach (array_keys($container->getServiceIdsForTag(ConsoleExtension::TAG_CONSOLE_COMMAND)) as $serviceId) {
                 $command = $container->get($serviceId);
-                $application->add($command);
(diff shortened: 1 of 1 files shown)


Jean85:
Because you're forcing Composer's SAT solver to use the beta; it seems that, if it's left with the choice, it prefers the stable tag.

We need Flex and its forced requirement.

Kocal:
Added Flex and updated the pinned version of `shivammathur/setup-php`. 
We now requires https://github.com/phpbench/container/pull/12 to be merged/released.

Jean85:
> container 2.2.3 was tagged.

Can you please restart the CI?

> still seems odd that Symfny seems to require a special tool to install it, given a choice I'd rather just `composer require symfony/console $symfonyVersion` in the pipeline. But it's fine to leave it as it is.

Flex does a lot more than that, because it locks down all Symfony components, so it influences indirect dependencies too. In the past I had issues without that because it can mix different major versions, leading to strange bugs in CI.

dantleech:
ok, looks like we've uncovered actual incompatibilit(y|ies) now

PR #1153, opened 2026-06-14
Expr: user defined functions and lambdas

dantleech:
This PR introduces a way to define lambdas in the expression language  and allows the definition of _user defined functions_:

```
(((foo, bar) => foo + bar)(1, 2) = 3) = true
```

While there is are no "higher order" functions that accept lambdas, this syntax is useful for _defining_ named functions:

```json
{
  "expression.user_functions": {
     "add_one": "(number) => number + 1"
  }
}
```

This is particular useful in report generation to avoid excessive repitition.

TODO:

- While users can register expressions, PHPBench itself can't currently and if it did it would contaminate the userspace. Perhaps the expression generator should be able to define it's own functions at "runtime" (?)


lib/Report/Generator/ExpressionGenerator.php
```diff
@@ -109,12 +114,30 @@ public function configure(OptionsResolver $options): void
                 ), E_USER_DEPRECATED);
             }
 
+            $baselineExpr = function (string $primaryExpr, string $baselineExpr) {
```
dantleech, line 117: avoiding this _would_ be the usecase here.

